PR #20, opened 2015-08-12
Add C wrapper

hominhquan:
-  Add C wrapper. User now can include clblast_c.h in C code.
-  StatusCode, Layout, Transpose etc. notation are used as-it in this C interface

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,7 +103,7 @@ set(ROUTINES ${LEVEL1_ROUTINES} ${LEVEL2_ROUTINES} ${LEVEL3_ROUTINES})
 # ==================================================================================================
 
 # Gathers all source-files
-set(SOURCES src/clblast.cc src/database.cc src/routine.cc src/utilities.cc)
+set(SOURCES src/clblast.cc src/database.cc src/routine.cc src/utilities.cc src/clblast_c.cc)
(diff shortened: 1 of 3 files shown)


hominhquan:
Please note that this is just a first version that compiles. I will write some test cases soon.

CNugteren:
Thanks for the contribution! A couple of remarks after a first quick scan of the merge request:
* Could you add your changes to the `development` branch instead and then create a new merge request into `CNugteren:development`? Thanks!
* Do you think we need test-cases? I imagine that the C wrapper will not change much after its first version. And it just calls the regular interface directly, so there is also not much that can go wrong.
* Why is the regular C++ header included in the C-version? Doesn't that defeat the purpose?
* Are the enum conversion functions really needed? Can't they automatically be converted using a regular cast since their underlying data-structure is the same?
* Why do you need the vector data-types? Perhaps `float2` and `double2` are useful for complex data-types, but the others?
* I'm not sure I like the pre-processor defines. Sure, they make the code more concise and remove code-duplication. However, they also make the code and potential compiler warnings/errors more unreadable.

I can also make a new branch and create a quick example (e.g. one routine) of what I had in mind, based on your initial version?

CNugteren:
I am closing this, since we now have a now have a basic C API in the `c_api` branch. It contains only a single routine so far, but I'm working on auto-generating the contents using a Python script.

Issue #33, opened 2016-03-30
Provide Visual Studio 2010 support (for python 3.4)


hughperkins:
Hi,

I have a couple of neural net projects for OpenCL, ie https://github.com/hughperkins/DeepCL and https://github.com/hughperkins/clnn  They're currently using clblas.  I'm tentatively interested in investigating using clblast instead, since eg [psyhtest](https://github.com/psyhtest) recommends it [here](https://github.com/amd/OpenCL-caffe/issues/17#issuecomment-203356483).  I notice you require c++11  I currently support C++0x and Visual Studio 2010 as standard across my projects, for a few reasons:
- many people still use old version of gcc, like 4.4 or so
- many people are using Python 3.4 (which is Visual Studio 2010), or Python 2.7 (which is Visual Studio 2008, but 2008 is basically approximately compatible with 2010, in either direction; both are approximately c++0x standard, in my experience, with a couple of smallish discrepancies).

I know that C++11 gives you a few conveniences as a developer, which it's hard to resist :-)  To what extent might you consider dropping your compiler requirements from c++11 down to c++0x /  Visual Studio 2010?

Hugh

hughperkins:
You know what?  I changed line 62 of CMakeLists.txt from `c++11` to `c++0x`, and the library built just fine, no errors :-)

(Mind you, I seem to be lacking tuners, and tests for some reason, with samples set to 'OFF'. Not sure why?)

hughperkins:
(Hmmm, the tests have clBLAS as a dependency? :-P)

CNugteren:
Good to hear! Be aware that CLBlast is still very much under development, so some features might be missing or performance might be sub-optimal. Perhaps it is better to wait 2 or 3 more months? In the meantime it is a good idea to see what is still missing, so I can focus on those issues.

The library is C++11, and it's going to stay that way. However, I am trying to use only those features supported earlier on by compilers. For example GCC 4.7 and MSVC 2015 are not fully C++11 compatible, but they do properly compile the code.

I don't think C++0x is something you can theoretically provide support for, because it is just an earlier development name for C++11. What could be done is trying older compilers and see what is already supported and what can be easily be fixed (e.g. by changing the compiler flag). So, I propose the following:
* You let me know which MSVC versions already work out-of-the-box (I don't have those installed), and we can update the README.
* I'll add testing on older GCC and Clang versions to my TODO list, trying to see what can be done there.

hughperkins:
[...]

PR #34, opened 2016-03-30
Downgrade compiler requirement from C++11 to C++0x, to enable compati…

hughperkins:
Downgrade compiler requirement from C++11 to C++0x, to enable compatibility with earlier versions of gcc, and with python 2.7 on Windows, and Python 3.4 on Windows

Addresses https://github.com/CNugteren/CLBlast/issues/33


CNugteren:
Unless you have a specific compiler in mind for which this works, I will close this pull request. Tested myself with GCC 4.6 and that doesn't seem to work: it is missing too many C++11 features.

hughperkins:
The specific compiler is: msvc 2008.  It's true that c++0x and msvc2008 are not terribly close though, so we could just look at ensuring that build on msvc2008 works.

hughperkins:
(or: at least, msvc2010; and I can build for msvc2008 in another way)

CNugteren:
Yeah, OK, but this single 'fix' doesn't actually fix compilation in Visual Studio 2008. So, I think we should continue this discussion in the other issue or PR related to actual changes to the code base. In that case, I would propose to try to go to Visual Studio 2013 first, since that one is already missing many many C++11 features (https://msdn.microsoft.com/en-us/library/hh567368.aspx). If that will ever work, we can go back in history one step at a time.

hughperkins:
> So, I think we should continue this discussion in the other issue or PR related to actual changes to the code base.

Yes thats fine for me

> I would propose to try to go to Visual Studio 2013 first,

vs2013 doesnt really help me much: my goal is to build for python 2.7 or 3.4, wihch is msvc2008 and msvc2010.

I have submitted an initial PR for how the changes for msvc2008/2010 will look.  They're quite radical, so not sure how willing you would be to accept those?

Issue #35, opened 2016-03-30
Ideally, tests should not depend on clBLAS library


hughperkins:
Ideally, tests should not depend on clBLAS library.

For example, you could either compare with atlas/openblas, or use some simple slow-but-probably-correct cpu code perhaps like:
```
#include <iostream>
#include <sys/types.h>
#include <stdio.h>
#include <string.h>
#include <clBLAS.h>
#include <stdlib.h>
using namespace std;

  cl_int err;
  cl_platform_id platform = 0;
  cl_device_id device = 0;
  cl_context_properties props[3] = { CL_CONTEXT_PLATFORM, 0, 0 };
  cl_context ctx = 0;
  cl_command_queue queue = 0;
  cl_mem bufA, bufB, bufC;
  cl_event event = NULL;
  int ret = 0;

void clgemm(int colmaj, char transAchar, char transBchar, int M, int N, int K, float alpha, float *A, int lda,
     float *B, int ldb, float beta, float *C, int ldc, float *result) {
clblasTranspose transA = transAchar == 'n' ? clblasNoTrans : clblasTrans;
clblasTranspose transB = transBchar == 'n' ? clblasNoTrans : clblasTrans;

size_t off = 0;
size_t offA = 0;
size_t offB = 0;
size_t offC = 0;

clblasOrder order;
if(colmaj == 1 ) {
  order = clblasColumnMajor;
} else {
  order = clblasRowMajor;
}

  bufA = clCreateBuffer(ctx, CL_MEM_READ_ONLY, M * K * sizeof(*A),
                        NULL, &err);
  bufB = clCreateBuffer(ctx, CL_MEM_READ_ONLY, K * N * sizeof(*B),
                        NULL, &err);
  bufC = clCreateBuffer(ctx, CL_MEM_READ_WRITE, M * N * sizeof(*C),
                        NULL, &err);

  err = clEnqueueWriteBuffer(queue, bufA, CL_TRUE, 0,
      M * K * sizeof(*A), A, 0, NULL, NULL);
  err = clEnqueueWriteBuffer(queue, bufB, CL_TRUE, 0,
      K * N * sizeof(*B), B, 0, NULL, NULL);
  err = clEnqueueWriteBuffer(queue, bufC, CL_TRUE, 0,
      M * N * sizeof(*C), C, 0, NULL, NULL);

  err = clblasSgemm(order, transA, transB, M - off, N - off, K - off,
                       alpha, bufA, offA, lda,
[...]

PR #38, opened 2016-03-30
add draft concept for testing xgemm correctness vs cpu

hughperkins:
add draft concept for testing xgemm correctness vs slow-but-probably-correct cpu routines, addresses https://github.com/CNugteren/CLBlast/issues/35

Results on my machine:
```
got platforms
1000 done
2000 done
3000 done
4000 done
5000 done
6000 done
7000 done
8000 done
9000 done
10000 done
11000 done
12000 done
13000 done
14000 done
15000 done
16000 done
All done! numTestsDone: 16384
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,13 +91,7 @@ if(TUNERS)
 endif()
 
 # Locates the clBLAS library in case the tests need to be compiled. "FindclBLAS.cmake" is included.
-if(TESTS)
-  find_package(clBLAS)
-  if(NOT CLBLAS_FOUND)
-    message(STATUS "Could NOT find clBLAS, disabling the compilation of the tests")
-    set(TESTS OFF)
-  endif()
-endif()
+find_package(clBLAS)
 
 # ==================================================================================================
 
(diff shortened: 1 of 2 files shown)


hughperkins:
Note however that times are different on clblast vs clblas 2.11.  For clblast:
```
time ./test_xgemm_vs_cpu 
...
All done! numTestsDone: 16384

real	0m19.151s
user	0m5.996s
sys	0m13.080s
```
For clblas 2.11:
```
$ time ./gemmtest >/dev/null

real	0m6.165s
user	0m2.264s
sys	0m3.804s
```

hughperkins:
(and both are built in 'Debug')

CNugteren:
Thanks for the contribution. However I am going to close this pull request, because:
* I don't think it is a good idea to write our own CPU BLAS, see #35.
* If we decide on writing our own BLAS, we should use the framework already in-place in CLBlast
* Committed code should follow the modern C++ style currently present in the library, and OpenCL API calls should be abstracted away with the `clpp11.h` header file.

Further discussion in the corresponding issue.

PR #39, opened 2016-03-31
Preliminary fixes for msvc10

hughperkins:
Some preliminary fixes for msvc10
- so you can see the types of things that arent working, and need to be modified, to work on msvc10
- so you can consider to what extent you can accept, or not, such changes

(Note that the CMP0054 changes in CMakeLists.txt are not essential for window build; they are warnings only)

Prototype for what would be required to address https://github.com/CNugteren/CLBlast/issues/33


CNugteren:
I'll close this per our discussion in #33. Similar changes are now also in the `visual_studio` branch, but I don't think that branch will remain alive. Visual Studio 2015 or pre-built binaries is the way to go here.

Issue #41, opened 2016-04-05
SGEMM failures in CLBlast tests


hughperkins:
Two failures in SGEMM test, using `-full_test`, on NVIDIA 940M, with libopenblas installed, on ubuntu 64-bit

```
* Testing 'regular behaviour' for '1 (col-major) 1 (transposed) 0 (regular)':
   ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
   ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
   ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
   ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
   ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
   ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
   ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
   ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
   ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
   ................................................................
   ................................................................
   ................................................................
   ................................................................
   ................................................................
   ........................................::::::::::::::::::::::::
   ::::::::::::::::::::::::::::::::::::::::::::::::................
   ........................................................::::::::
   ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
   ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
   ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
   ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
   ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
   ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
   ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
   ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
   ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
   ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
   ................................................................
   ................................................................
   ................................................................
   ................................................................
   ................................................................
   ........................................::::::::::::::::::::::::
   ::::::::::::::::::::::::::::::::::::::::::::::::................
[...]

PR #57, opened 2016-05-05
Locate the C BLAS library before the F77 one.

psyhtest:
On my Chromebook 2 with Gentoo Linux, the following packages are available:
```
*  sci-libs/blas-reference
      Latest version available: 20070226-r4
      Latest version installed: 20070226-r4
      Size of files: 5208 KiB
      Homepage:      http://www.netlib.org/blas/
      Description:   Basic Linear Algebra Subprograms F77 reference implementations
      License:       BSD

*  sci-libs/cblas-reference
      Latest version available: 20030223-r6
      Latest version installed: 20030223-r6
      Size of files: 194 KiB
      Homepage:      http://www.netlib.org/blas/
      Description:   C wrapper interface to the F77 reference BLAS implementation
      License:       public-domain
```
I have changed the order of the first two library names in `cmake/Modules/FindCBLAS.cmake`:
```
-  NAMES blas cblas mkl blis openblas accelerate
+  NAMES cblas blas mkl blis openblas accelerate
```
This  allows CMake to locate the reference C BLAS library (`/usr/lib/libcblas.so`) before the F77 one (`/usr/lib/libblas.so`), and thus solves the linking issue I reported in #41 (for which I should now be able to provide more information).

### DIFF
--- a/cmake/Modules/FindCBLAS.cmake
+++ b/cmake/Modules/FindCBLAS.cmake
@@ -48,7 +48,7 @@ mark_as_advanced(CBLAS_INCLUDE_DIRS)
 
 # Finds the library
 find_library(CBLAS_LIBRARIES
-  NAMES blas cblas mkl blis openblas accelerate
+  NAMES cblas blas mkl blis openblas accelerate
   HINTS ${CBLAS_HINTS}
   PATH_SUFFIXES
     lib lib64 lib/x86_64 lib/x64 lib/x86 lib/Win32 lib/import lib64/import


blueberry:
you should be able to set this from the environment with $CBLAS_ROOT. See the README.

psyhtest:
The issue here is not with the library directory path (both the C and F77 libraries are in /usr/lib/) but with the library name.

CNugteren:
Thanks, that makes sense. I'll merge it in.

Issue #71, opened 2016-06-30
Failing with CL_INVALID_EVENT_WAIT_LIST


gcp:
I'm getting a "kKernelLaunchError" status code when launching my kernels. Adding a printf in the CheckError function to see the real underlying error, I'm getting a -57, i.e. CL_INVALID_EVENT_WAIT_LIST.

    const size_t M = outputs;         == 96
    const size_t N = spatial_out;    == 361
    const size_t K = filter_dim;       == 550

    auto queue_plain = queue();
    auto status = clblast::Gemm(
        clblast::Layout::kRowMajor,
        clblast::Transpose::kNo, clblast::Transpose::kNo,
        M, N, K,
        1.0f,
        weights[0](), 0, filter_dim,
        bufferCol(), 0, spatial_out,
        0.0f,
        bufferOutput(), 0, spatial_out,
        &queue_plain);

Looks like an issue with pad/transpose handling?

gcp:
For what it's worth, Release 0.7.1 works correctly, so this is a regression.

CNugteren:
Thanks for reporting. I am unable to reproduce the issue on my system though. Couple of questions:

1. Does the included `samples/sgemm.cpp` work for you? And what if you modify it to use your M, N, K dimensions and alpha/beta settings? And what if you remove the optional `event` argument? Those were the steps I did to try to reproduce the issue on my system anyway.
2. Even better - if you have the time - can you compile and run the included GEMM tests? See the README for instructions on how to build the tests.
3. Does the error also occur if you supply an event yourself (it is an option argument to the GEMM call)?
4. What hardware are you running on? Perhaps the issue is vendor specific?

gcp:
1) No, I get "Completed SGEMM in 241.348 ms with status -2048". Removing the event argument does not change this. I tried something similar with my own code, which is why I think the event problem occurs when the pad/transpose kernels are added internally in clblast.
2) I can't get it to recognize my OpenBLAS installation. I'm compiling on MSVC2015 in 32-bit mode which (I'm guessing) isn't the most tried and tested environment. (Compiling TUNERS also fails due to symbols not found in the clblast lib).
3) Yes, I tried this, with no effect. Again, I think the event issue is in clblasts own handling of sgemm.
[...]

PR #72, opened 2016-06-30
Ensure the clGetKernelWorkGroupInfo return value fits.

gcp:
In LocalMemUsage(), there's a first call to clGetKernelWorkGroupInfo to get the "bytes" amount needed to store the result from CL_KERNEL_LOCAL_MEM_SIZE. However, the actual value passed is an "auto result = size_t", which in 32-bit mode is 4 bytes, regardless of the previous return value. The spec describes that this is actually a cl_ulong which is 8 bytes. So at this point clGetKernelWorkGroupInfo will corrupt the stack because the passed variable can't actually contain the result it wants to write. 

Fixes issue #71


CNugteren:
OK, thanks! That seems like a bug indeed, I haven't tested on 32-bit, so thanks for doing that. Two comments:

1. Instead of `cl_ulong result = size_t{0};` I would suggest `auto result = cl_ulong{0};` to make it a bit more understandable for the reader of the code and in line with the code style.
2. If the result is indeed an `unsigned long`, perhaps we should also return it from the function as that type. Now we could lose information by casting it to a `size_t`. But I guess it's fine because this shouldn't exceed the max for a 32-bit value anyways.

Could you fix point 1? And then also create a new pull request to merge it into the `development` branch instead of into `master`? Thanks! I can also fix it myself but this way your contribution will be official :-)

gcp:
Sounds fine. I considered (2) initially but wasn't sure if I wanted to adjust all callers. But if you prefer that, I'll make that change too.

CNugteren:
No, I guess just (1) is fine.

gcp:
Looks like there's only a single user of the return value, receiving it into an "auto". That's then used to call another function...that has a cl_ulong and static_casts it to size_t to avoid a type mismatch. So (2) actually makes everything simpler :)

gcp:
Looks like github doesn't allow changing the target of a pull request. Huh! I'll close and reopen for the development branch.

PR #80, opened 2016-07-13
Make sure the passed types are large enough.

gcp:
Make sure all out parameters that are passed to functions such as clGetDeviceInfo are large enough to contain the replies.

Also make the default of GetInfo an ulong, though it's worth considering removing the non-templated function here, as it's prone to exactly these kind of errors.


CNugteren:
Thanks for the PR. I'll remove the non-templated function as you suggested, it makes sense.

PR #81, opened 2016-07-13
Groundwork for device-specific routines

intelfx:
This pull request lays a bit of groundwork for adding support and infrastructure for device-specific routines and kernels in CLBlast (as discussed with @psyhtest and exemplified by the [Mali-optimized sgemm routine](http://malideveloper.arm.com/downloads/OpenCL/example_codes/sgemm/)).

Right now these changes facilitate creating "overlay" device-specific libraries (which are linked on top of CLBlast and implement a subset of its routines with same API/ABI). An example of such library can be seen in [intelfx/CLBlast#mali-overlay](https://github.com/intelfx/CLBlast/tree/mali-overlay).

Doing these things is certainly possible without merging this PR ([intelfx/CLBlast#mali-hackish](https://github.com/intelfx/CLBlast/tree/mali-hackish) links with unmodified CLBlast master), but it required me to ~~add insult to injury~~violate ODR a couple of times in an unsafe way by monkey-patching existing CLBlast classes, so I thought I'd better factor out the changes to existing API, so that overlay library will contain only the routine class itself.

Actually, this is not final, so this may or may not be merged. E. g. the xgemm routine changes are there just to make it easier to modify that function for device-specific needs, and similar generalizations probably should be applied to every routine out there. Also there are a couple of unrelated (yet useful) changes.

In the near future, I intend to expand this into a full-featured API for replacing parts of CLBlast in runtime without playing games with the linker and happily violating ODR in the process.


CNugteren:
That sounds interesting, thanks for your work! Indeed, this should then be applied to every routine, not just GEMM.

Indeed, I tried to optimise the CLBlast kernels for ARM Mali, but the ARM OpenCL compiler doesn't handle the style of kernel programming used. For example, inside the kernel I use these kinds of structures:
```
#define WPT 4 // or some other value: 1, 2, 4, 8, ...
float values[WPT]
for (int i = 0; i < WPT; ++i) { values[i] = 0.0; }
```
I expect this array and this loop to be 'unrolled' as registers, but this compiler cannot do that, adding a lot of loads/stores and loop overhead/indexing. Anyway, that's a different discussion.

About the PR:

1. Can you split it up in smaller chunks? For example the improvements to `src/clpp11.hpp` are separate right? With smaller PRs its easier for me to understand what you added and easier to merge things in.
2. Could you make the PR into the `development` branch instead of `master`?
3. There is some linker error on OS X for `unsigned long clblast::ConvertArgument<unsigned long>(char const*)`, see https://travis-ci.org/CNugteren/CLBlast/jobs/144547014

intelfx:
Unfortunately I can't change the base branch for the PR, so I'll reopen this against the correct branch.

PR #84, opened 2016-07-22
Groundwork for device-specific routines (retry)

intelfx:
This pull request lays a bit of groundwork for adding support and infrastructure for device-specific routines and kernels in CLBlast (as discussed with @psyhtest and exemplified by the [Mali-optimized sgemm routine](http://malideveloper.arm.com/downloads/OpenCL/example_codes/sgemm/)).

Right now these changes facilitate creating "overlay" device-specific libraries (which are linked on top of CLBlast and implement a subset of its routines with same API/ABI). An example of such library can be seen in [intelfx/CLBlast#mali-overlay](https://github.com/intelfx/CLBlast/tree/mali-overlay).

Doing these things is certainly possible without merging this PR ([intelfx/CLBlast#mali-hackish](https://github.com/intelfx/CLBlast/tree/mali-hackish) links with unmodified CLBlast master), but it required me to ~~add insult to injury~~violate ODR a couple of times in an unsafe way by monkey-patching existing CLBlast classes, so I thought I'd better factor out the changes to existing API, so that overlay library will contain only the routine class itself.

Actually, this is not final, so this may or may not be merged. E. g. the xgemm routine changes are there just to make it easier to modify that function for device-specific needs, and similar generalizations probably should be applied to every routine out there. Also there are a couple of unrelated (yet useful) changes.

In the near future, I intend to expand this into a full-featured API for replacing parts of CLBlast in runtime without playing games with the linker and happily violating ODR in the process.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,7 +95,7 @@ set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${CFLAGS}")
 # ==================================================================================================
 
 # Package scripts location
-set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules/")
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${clblast_SOURCE_DIR}/cmake/Modules/")
 
 # Requires OpenCL. It is found through the included "FindOpenCL.cmake" in CMAKE_MODULE_PATH.
 find_package(OpenCL REQUIRED)
--- a/src/clpp11.hpp
+++ b/src/clpp11.hpp
@@ -109,7 +109,9 @@ class Event {
 
   // Accessor to the private data-member
   cl_event& operator()() { return *event_; }
+  const cl_event& operator()() const { return *event_; }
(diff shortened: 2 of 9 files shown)


intelfx:
I had to recreate this PR against a different branch, so I've lost the first round of review. Here it is:

> Can you split it up in smaller chunks? For example the improvements to src/clpp11.hpp are separate right? With smaller PRs its easier for me to understand what you added and easier to merge things in.

The logically unrelated changes are in separate commits, so you can review them per-commit on the "Commit" tab... Or do you really want me to make each of the commits an independent PR?

> Could you make the PR into the development branch instead of master?

Done.

> There is some linker error on OS X for unsigned long clblast::ConvertArgument<unsigned long>(char const*), see https://travis-ci.org/CNugteren/CLBlast/jobs/144547014

TBH, I could not understand where this error comes from. Everything seemed correct. So I just dropped the relevant commit.

CNugteren:
Thanks for your work! I could indeed nicely evaluate the commits one by one. I merged in your changes and afterwards made some minor improvements, mostly to the comments. Let me know if you need any further help with understanding the CLBlast structure for your work.

Issue #85, opened 2016-07-24
Adding installable `find_package` file (for dependent projects) and bumping CMake dependency


intelfx:
Hello,

I'm trying to make it easier to link to CLBlast from dependent projects (say Caffe). The typical way of writing a module script for `find_package()` seems like too much boilerplate for my idealistic mind, so I'm trying to use the "config" mode of find_package (second signature in the doc), that is, the `install(EXPORT ... DESTINATION ...)` command.

(It installs a declarative-like target description file with absolute pathes to the installed files, which may then be searched by the dependent project's CMakeLists.txt and imported to create an "imported target". This is very similar to pkg-config.)

The problem is that exporting targets needs at least CMake version 3.2. Can we bump the requirement from 2.8 to 3.2?

(The current Ubuntu 16.04 LTS has CMake 3.2 in base image and 3.5 in updates.)

CNugteren:
That could be a nice addition indeed. But going to CMake 3.2 from 2.8 is quite a big jump in requirements to be honest. Can we somehow guard the relevant part such that it is only executed if the newest CMake is available?

By the way, can you point me to the docs of this specific feature (and it's dependency on 3.2). Thanks!

intelfx:
As I've just said in the PR, I've misread the documentation three times in a row (`install(EXPORT ...)` vs `export(EXPORT ...)` vs `install(TARGETS ... EXPORT ...)`) and actually there is no change to CMake dependency.

CNugteren:
Thanks again for improving the CMake infrastructure. I've accepted your PR and tested on CMake 3.0 and 3.5, both seem to work fine. I couldn't find a system with 2.8.11 at this moment. All targets build and run on my system, and indeed `CLBlastConfig.cmake` is now installed. Don't know how to test if that has proper contents though, I'll leave that up to you :-)

PR #86, opened 2016-07-25
CMakeLists.txt: provide a find_package() config for dependent projects

intelfx:
I believe in pipelining, so here it is, in case you are OK with bumping the CMake dep to 3.2 (see #85).

This allows me to link to CLBlast in Caffe's CMakeLists.txt in ~15 lines of code (see [intelfx/caffe#clblast](https://github.com/intelfx/caffe/tree/clblast)) without writing any boilerplate, with support for non-standard prefixes and all that (so I'm not NOT just adding `-lclblast` to the compiler command line and calling it a day).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,11 +127,6 @@ endif()
 
 # ==================================================================================================
 
-# Includes directories: CLBlast and OpenCL
-include_directories(${clblast_SOURCE_DIR}/include ${clblast_SOURCE_DIR}/src ${OPENCL_INCLUDE_DIRS})
-
-# ==================================================================================================
-
 # Sets the supported routines and the used kernels. New routines and kernels should be added here.
 set(KERNELS copy_fast copy_pad transpose_fast transpose_pad xaxpy xdot xger xgemm xgemv)
 set(SAMPLE_PROGRAMS_CPP sgemm)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -173,17 +168,27 @@ endforeach()
 add_library(clblast SHARED ${SOURCES})
 target_link_libraries(clblast ${OPENCL_LIBRARIES})
 
+# Includes directories: CLBlast and OpenCL
+target_include_directories(clblast PUBLIC
+                           $<BUILD_INTERFACE:${clblast_SOURCE_DIR}/include>
+                           $<BUILD_INTERFACE:${clblast_SOURCE_DIR}/src>
+                           $<INSTALL_INTERFACE:include>
+                           ${OPENCL_INCLUDE_DIRS})
+
```
CNugteren: This is still CMake 2.8.11 if I am correct, so it would be good to have this in anyway. But then something like this perhaps? Or am I missing something?
```
target_include_directories(clblast
                           PRIVATE
                           $<BUILD_INTERFACE:${clblast_SOURCE_DIR}/src>
                           PUBLIC
                           $<BUILD_INTERFACE:${clblast_SOURCE_DIR}/include>
                           $<INSTALL_INTERFACE:include>
                           ${OPENCL_INCLUDE_DIRS})
```

Also, then I think it is best to also use `target_include_directories` for the remainder of the targets (tests, samples, clients, tuners) further down in the CMake file.

CMakeLists.txt
```diff
@@ -173,17 +168,27 @@ endforeach()
...
 
 # Installs the library
-install(TARGETS clblast DESTINATION lib)
+install(TARGETS clblast EXPORT CLBlast DESTINATION lib)
 install(FILES include/clblast.h DESTINATION include)
 install(FILES include/clblast_c.h DESTINATION include)
 install(FILES include/clblast_half.h DESTINATION include)
 
+# Installs the config for find_package in dependent projects
+install(EXPORT CLBlast DESTINATION lib/cmake/CLBLast FILE CLBlastConfig.cmake)
```
CNugteren: Is this the only CMake 3.2 specific line? Or is line 184 also not supported in 2.8.11?

CMakeLists.txt
intelfx: > This is still CMake 2.8.11 if I am correct, so it would be good to have this in anyway.

AFAICS, yes.

> But then something like this perhaps? Or am I missing something?

I don't think so; the tests and remaining auxiliary targets use "private" headers from `src/`, like `"utilities.hpp"`.

> Also, then I think it is best to also use target_include_directories for the remainder of the targets (tests, samples, clients, tuners) further down in the CMake file.

I'll do this if you want me to. I was trying to avoid unrelated changes.

CMakeLists.txt
intelfx: Yes, it is. Previously there was `export(TARGETS ...)`, but the documentation explicitly says against using it for creating installable file (as opposed to creating files internal to the build process).

intelfx:
BTW, CMake 3.2.2 in the Travis builder still doesn't like something against the `install(EXPORT ...)` line. I'll try to see what's wrong.

intelfx:
OK, for the test failures, can we relocate OpenCL installation prefix (`${OPENCL_ROOT}`) somewhere out of the CLBlast's build prefix? CMake seems to dislike this.

CMakeLists.txt
intelfx: In fact, it is CMake 3.0, not 3.2.

intelfx:
@CNugteren: with changes to CI scripts, the tests now pass and the CMake requirement is actually 3.0, not 3.2. Does this change your opinion?

intelfx:
Damn. I'm misreading the documentation for the third time in row. `install(EXPORT ...)` exists in 2.8.11. Sorry for misleading you; there should be no changes to CMake dependency.

CNugteren:
OK, that's good news! I agree that the CMake documentation can be quite confusing. I'll merge it in and verify if everything still works on a system with older CMake variants installed.

Issue #88, opened 2016-08-09
Does CLBlast supports opencl 1.1 EP


Mezzano:
I'm wondering if the Embedded Profile is supported as I get a [floating point exception](https://github.com/rickyHong/caffe-for-clblast-branch/issues/1) when trying to run in GPU mode.

CNugteren:
I am not sure if there are any limiting features in CLBlast. Could you run the included tests and tell me which ones fail/succeed?

dvasschemacq:
Hi, I'm working with Mezzano, we were able to run the test now, we tried this one : clblast_test_xsyr

And the output is `* Running on OpenCL device 'Vivante OpenCL Device'.
* Starting tests for the 'SSYR' routine. Legend:
   : -> Test produced correct results
   . -> Test returned the correct error code
   X -> Test produced incorrect results
   / -> Test returned an incorrect error code
   \ -> Test not executed: OpenCL-kernel compilation error
   o -> Test not executed: Unsupported precision
   - -> Test not completed: Reference CBLAS doesn't output error codes
* Testing 'regular behaviour' for '101 (row-major) 121 (upper)':
   OpenCL compiler error/warning: (358:0) : error : syntax error at '__kernel'

-OpenCL compiler error/warning: (358:0) : error : syntax error at '__kernel'

-OpenCL compiler error/warning: (358:0) : error : syntax error at '__kernel'

-OpenCL compiler error/warning: (358:0) : error : syntax error at '__kernel'

-OpenCL compiler error/warning: (358:0) : error : syntax error at '__kernel'

....
....
....

   Pass rate   0.0%: 0 passed / 12 skipped / 0 failed
* Completed all test-cases for this routine. Results:
   0 test(s) passed
   48 test(s) skipped
   0 test(s) failed

* Running on OpenCL device 'Vivante OpenCL Device'.
* Starting tests for the 'DSYR' routine.
* All tests skipped: Unsupported precision

* Running on OpenCL device 'Vivante OpenCL Device'.
* Starting tests for the 'HSYR' routine.
* All tests skipped: Unsupported precision
`

So there seems to be a precision problem that can come from the EP we have. 

But will it block the working of the library, I don't think, normally it should work but with less precision due to our hardware? Is it right?

CNugteren:
[...]

PR #89, opened 2016-08-18
Adapt opencl files for 1.1 OpenCL

dvasschemacq:
In OpenCL 1.1 __kernel has to be before __attribute__, at least with
Vivante compiler.


CNugteren:
Thanks, this change looks fine (as discussed in #88). However, can you make this pull request to the `development` branch instead? You'll have to re-create it unfortunately.

bhack:
Is it still required? See https://github.com/blog/2224-change-the-base-branch-of-a-pull-request

CNugteren:
I followed bhack's advice, but now the pull request is not as clean as before unfortunately. I'll have to merge it in later, probably tomorrow.

dvasschemacq:
Ok, if we have another modification we will push it on the development branch next time. :)

No more action required from my side as I understand?

PR #92, opened 2016-08-25
test/correctness: read platform and device from environment

intelfx:
Support passing environment variables `CLBLAST_PLATFORM` and `CLBLAST_DEVICE` instead of `-platform` and `-device` arguments to test executables.
    
This is for `ctest`.

(If you remember, earlier I tried to include this commit but it failed unit tests. Now I finally understood the problem.)


intelfx:
Ugh, wrong branch...

PR #93, opened 2016-08-25
test/correctness: read platform and device from environment

intelfx:
Support passing environment variables `CLBLAST_PLATFORM` and `CLBLAST_DEVICE` instead of `-platform` and `-device` arguments to test executables.
    
This is for `ctest`.

(If you remember, earlier I tried to include this commit but it failed unit tests. Now I finally understood the problem.)

### DIFF
--- a/src/utilities.cpp
+++ b/src/utilities.cpp
@@ -161,6 +161,8 @@ template <typename T>
 T ConvertArgument(const char* value) {
   return static_cast<T>(std::stoi(value));
 }
+template size_t ConvertArgument(const char* value);
+
 template <> half ConvertArgument(const char* value) {
   return FloatToHalf(static_cast<float>(std::stod(value)));
 }
@@ -178,6 +180,15 @@ template <> double2 ConvertArgument(const char* value) {
   auto val = static_cast<double>(std::stod(value));
   return double2{val, val};
 }
+
+// Variant of "ConvertArgument" with default values
+template <typename T>
+T ConvertArgument(const char* value, T default_value) {
+
+  if (value) { return ConvertArgument<T>(value); }
+  return default_value;
+}
+template size_t ConvertArgument(const char* value, size_t default_value);
 
 // This function matches patterns in the form of "-option value" or "--option value". It returns a
 // default value in case the option is not found in the argument string.
--- a/src/utilities.hpp
+++ b/src/utilities.hpp
@@ -187,6 +187,10 @@ std::string ToString(T value);
 template <typename T>
 T ConvertArgument(const char* value);
 
+// Variant of "ConvertArgument" with default values
+template <typename T>
+T ConvertArgument(const char* value, T default_value);
+
 // Basic argument parser, matching patterns in the form of "-option value" and "--option value"
 template <typename T>
 T GetArgument(const int argc, char **argv, std::string &help,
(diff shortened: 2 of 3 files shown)


test/correctness/tester.cpp
```diff
@@ -27,8 +27,8 @@ template <typename T, typename U>
 Tester<T,U>::Tester(int argc, char *argv[], const bool silent,
                     const std::string &name, const std::vector<std::string> &options):
     help_("Options given/available:\n"),
-    platform_(Platform(GetArgument(argc, argv, help_, kArgPlatform, size_t{0}))),
-    device_(Device(platform_, GetArgument(argc, argv, help_, kArgDevice, size_t{0}))),
+    platform_(Platform(GetArgument(argc, argv, help_, kArgPlatform, ConvertArgument(getenv("CLBLAST_PLATFORM"), size_t{0})))),
+    device_(Device(platform_, GetArgument(argc, argv, help_, kArgDevice, ConvertArgument(getenv("CLBLAST_DEVICE"), size_t{0})))),
```
CNugteren: Where is `getenv` defined? Perhaps better to make it `std::getenv` and include `<cstdlib>` to be on the safe side (http://en.cppreference.com/w/cpp/utility/program/getenv).

src/utilities.cpp
```diff
@@ -157,10 +157,6 @@ std::string ToString(Precision value) {
 // Helper for the below function to convert the argument to the value type. Adds specialization for
 // complex data-types. Note that complex arguments are accepted as regular values and are copied to
 // both the real and imaginary parts.
-template <typename T>
-T ConvertArgument(const char* value) {
-  return static_cast<T>(std::stoi(value));
-}
```
CNugteren: Why move this to the header? Because of the template otherwise not being instantiated? That makes sense, but the rest of CLBlast contains only declarations in the headers and the implementation + optional explicit instantiation in the .cpp file. Perhaps it is nicer to keep it here and than explicitly instantiate it.

src/utilities.hpp
```diff
@@ -183,9 +183,19 @@ std::string ToString(T value);
...
+  return static_cast<T>(std::stoi(value));
+}
+
+// Variant of "ConvertArgument" with default values
+template <typename T>
+T ConvertArgument(const char* value, T default_value) {
+
+  if (value) { return ConvertArgument<T>(value); }
+  return default_value;
+}
```
CNugteren: See above: can we move this to the .cpp file? And then add this to explicitly instantiate it for `size_t` (which is what you need for the platform/device settings - see below):
```
template size_t ConvertArgument(const char* value, size_t default_value);
```

test/correctness/tester.cpp
intelfx: Thanks, fixed.

src/utilities.cpp
intelfx: I thought that this helper is small enough to put it in the header. OK though.

src/utilities.hpp
intelfx: This is even more helper-ish, but well.

CNugteren:
Thanks for the changes. I agree that some things could go (or better go) in the header, but I think here it is better to go for consistency - makes it easier to find where functions are.

PR #101, opened 2016-09-24
Add path to ref library header when building tests.

psyhtest:
Building the `test_correctness_common` target fails when a reference library header (e.g. `clBLAS.h`) is not in standard paths. This PR adds the header path to the target's include directories.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -356,7 +356,7 @@ if(TESTS)
                 test/correctness/tester.cpp test/correctness/testblas.cpp)
     target_include_directories(test_correctness_common PUBLIC
                                $<TARGET_PROPERTY:clblast,INTERFACE_INCLUDE_DIRECTORIES>
-                               ${clblast_SOURCE_DIR})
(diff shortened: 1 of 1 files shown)


CNugteren:
Yes, you are right, that one was missing, thanks!

I now see that the Windows builds are failing but that's unrelated to this PR as far as I can see it. I'll investigate that later.

psyhtest:
@CNugteren Yes, I noticed Windows builds failing yesterday, but I thought you were aware of the issues...

CNugteren:
Not sure why the builds are failing though. It seems AppVeyor upgraded from CMake 3.5 and 3.6 and now it complains:
```
CMake Error in CMakeLists.txt:
  Target "clblast" INTERFACE_INCLUDE_DIRECTORIES property contains path:
    "C:/projects/clblast/bin/opencl/include"
  which is prefixed in the source directory
```
But my AppVeyor sources are in `C:\projects\clblast\bin\clblast` so I don't understand why this happens. Anyway, I'll figure it out - it doesn't seem too urgent.

PR #103, opened 2016-09-26
Link clBLAS together with pthread

psyhtest:
Linking CLBlast tests with clBLAS as the reference library failed on an HP Z640 running Ubuntu 16.04 with messages similar to https://github.com/clMathLibraries/clBLAS/issues/154. The proposed fix is also similar.


CNugteren:
Two questions:

1. Shouldn't clBLAS itself already link to pthread? Aren't we fixing a clBLAS issue here within CLBlast?
2. It is probably more appropriate to add pthread support like this (hopefully that will be cross platform):`find_package(Threads)` and `set(REF_LIBRARIES ${REF_LIBRARIES} ${CLBLAS_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT})`

What do you think?

psyhtest:
> Shouldn't clBLAS itself already link to pthread? Aren't we fixing a clBLAS issue here within CLBlast?

Quite possibly :). However, it seems that clBLAS is slowly dying, with the latest release over 9 months ago. Even if someone fixes the issue and sends a PR, it might never get merged. And even if this issue does get fixed in a future release, we still won't be able to test CLBlast vs clBLAS on OpenCL 1.1 platforms (as the latest clBLAS release to support OpenCL 1.1 was 2.4). So I would add this just to ensure that CLBlast gets tested without extra hassle on as many platforms as possible.

> It is probably more appropriate to add pthread support like this (hopefully that will be cross platform)

I've tested the fix you proposed on the same platform, and it works! I don't know how clBLAS handles threading  on Windows (or whether it even works on Windows), but it's probably a good idea not to request `pthread` directly.

CNugteren:
That makes sense, thanks for the explanation.

PR #112, opened 2016-10-13
Add option to build shared or static library

shehzan10:
Tested this on Linux.

PR #115, opened 2016-10-14
Fixes for static lib compilation on Windows

shehzan10:
Missed testing on Windows when sending in #112.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,6 +65,15 @@ elseif(MSVC)
   endif()
 endif()
 
+# DLL Settings
+if(MSVC)
+  if(BUILD_SHARED_LIBS)
+    add_definitions(" /DCLBLAST_DLL")
+  else(BUILD_SHARED_LIBS)
+    add_definitions(" /DCLBLAST_STATIC")
+  endif(BUILD_SHARED_LIBS)
+endif(MSVC)
+
 # C++ compiler settings
 if(MSVC)
   set(FLAGS "/Ox")
@@ -195,7 +204,9 @@ target_include_directories(clblast PUBLIC
 
 # Sets the proper __declspec(dllexport) keyword for Visual Studio when the library is built
 if(MSVC)
-  target_compile_definitions(clblast PRIVATE COMPILING_DLL=1) # requires at least CMake 2.8.11
+  if(BUILD_SHARED_LIBS)
+    target_compile_definitions(clblast PRIVATE COMPILING_DLL=1) # requires at least CMake 2.8.11
+  endif(BUILD_SHARED_LIBS)
 endif()
 
 # Installs the library
--- a/include/clblast.h
+++ b/include/clblast.h
@@ -28,10 +28,14 @@
 // Exports library functions under Windows when building a DLL. See also:
 // https://msdn.microsoft.com/en-us/library/a90k134d.aspx
 #ifdef _WIN32
-  #ifdef COMPILING_DLL
-    #define PUBLIC_API __declspec(dllexport)
-  #else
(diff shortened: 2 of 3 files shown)


CNugteren:
Although I merged all of this in (#112 and #115), I am still a bit confused by the logic used here. First of all I see:
```
if(MSVC)
  if(BUILD_SHARED_LIBS)
    add_definitions(" /DCLBLAST_DLL")
  else(BUILD_SHARED_LIBS)
    add_definitions(" /DCLBLAST_STATIC")
  endif(BUILD_SHARED_LIBS)
endif(MSVC)
```
But I can't find any reference to `CLBLAST_STATIC` in the library, because this was optimised away by me in commit #53deed2 . Hopefully this is correct. If so, I believe this can be simplied to:
```
if(MSVC)
  if(BUILD_SHARED_LIBS)
    add_definitions(" /DCLBLAST_DLL")
  endif(BUILD_SHARED_LIBS)
endif(MSVC)
```
But then later on I see something very similar:
```
if(MSVC)
  if(BUILD_SHARED_LIBS)
    target_compile_definitions(clblast PRIVATE COMPILING_DLL=1)
  endif(BUILD_SHARED_LIBS)
endif()
```
So basically both `CLBLAST_DLL` and `COMPILING_DLL` are set for a shared MSVC build. Thus the `COMPILING_DLL` version can be removed and the actual code simplifies back to a version very close what it was before #b330ab086640382157688ea6b9633b5f0a22dac3:
```
#if defined(_WIN32) && defined(CLBLAST_DLL)
    #define PUBLIC_API __declspec(dllexport)
#else
  #define PUBLIC_API
#endif 
```

So is `__declspec(dllimport)` actually not needed, or am I missing something here?

shehzan10:
So there are 4 possibilities of the header file's inclusion on Windows:
1. Compiling DLL from source
2. Being included in an application when the library is a DLL
3. Compiling static lib from source
4. Being included in an application when the library is a static lib

Cases 1 and 2:
So when we are using shared libs, CLBLAST_DLL is defined which requires `__declspec(dllexport)` to be defined for case 1 and `__declspec(dllimport)` to be defined for case 2. This distinction is made by the `COMPILING_DLL` definition which is only defined for the library but not for the examples, tests etc. Which means when `COMPILING_DLL` is enabled, the library is being compiled, and when it is not defined, the library is being used.

Cases 3 and 4:
When compiling or using static libs, `__declspec` should not be used in either case. Hence is static lib is defined, then `PUBLIC_API` is empty, as is the case when Windows is not used.

There are alternate ways to arrange the if block if you so wish, but these cases are definitely required.

CNugteren:
OK, thanks for clarifying. My mistake in the reasoning above was that `COMPILING_DLL` is always set when `CLBLAST_DLL` is set and vice versa. The vice versa part doesn't apply indeed, because `COMPILING_DLL` is target-specific (only library, not for the tests/samples/clients as you mentioned).

Then what is currently implemented is indeed correct:
```
if(MSVC)
  if(BUILD_SHARED_LIBS)
    add_definitions(" /DCLBLAST_DLL")
  endif()
endif(MSVC)

if(MSVC)
  if(BUILD_SHARED_LIBS)
    target_compile_definitions(clblast PRIVATE COMPILING_DLL=1)
  endif(BUILD_SHARED_LIBS)
endif()
```

and

```
#if defined(_WIN32) && defined(CLBLAST_DLL)
  #if defined(COMPILING_DLL)
    #define PUBLIC_API __declspec(dllexport)
  #else
    #define PUBLIC_API __declspec(dllimport)
  #endif
#else
  #define PUBLIC_API
#endif
```

Issue #110, opened 2016-10-11
Exceptions and error handling model inside the library


intelfx:
I've noticed that the exception model inside the library is pretty strange. The OpenCL wrapper throws exceptions which are caught halfway by nested catch statements repeated inside each routine and converted to error codes.

At the same time, the exception thrown from `database.cpp` is not caught in any way (and I vaguely remember other exceptions leaking into the client, too).

Will you accept a patchset converting the library to use custom exception objects carrying their own own error codes, and then catching all exceptions at the topmost level only?

CNugteren:
Perhaps that is a good idea indeed. The library was not really designed properly in terms of exception handling I guess. Perhaps this is also partly due to the fact that it should return error codes in the end and not throw exceptions. Or could we have both: error codes for expected errors (e.g. wrong input argument values) and exceptions for internal errors / bugs in the library (e.g. database errors)?

Also the reason that the OpenCL wrapper is different is that it is also used in an external project (CLCudaAPI).

But if you have a good plan and good arguments I will accept changes to the exception model.

blueberry:
Please also note that CLBlast is used in external libraries on JVM (JOCLBlast and Neanderthal), so do not break the existing way we can catch those errors :)

intelfx:
@blueberry Which errors do you handle and how exactly do you want to be able to handle them?

Current model of error handling in CLBlast is really inconsistent, e. g. any (both OpenCL, internal CLBlast and C++ runtime) errors thrown in "setup" phase are not handled at all, but any errors thrown in "run" phase (inside of the Routine subclasses) are wildcard caught, erased of actual exception type(!) and returned as generic error codes coarsely corresponding to the high-level action that caused an error.

blueberry:
Probably the easiest and most consistent way for managd environments such as JVM is "the C way", just like in plain OpenCL. You call the method, and it returns the error code that you handle as you please in that environment. Of course, I don't mind if there is something else for C++, but that would probably be a bit more difficult to handle from java. I think @gpu is the right person to ask about those things, since he created the JVM CLBlast wrapper.

CNugteren:
@intelfx You are right that it is a mess. I propose the following changes:

1. All exceptions thrown internally in CLBlast should be caught and a proper error code should be returned instead. If an appropriate error code does not exist yet it should be added. For JOCLBlast this would mean just adding some error codes to the enum.

[...]

PR #117, opened 2016-10-18
Convert to use C++ exceptions internally

intelfx:
Refactor the whole codebase to only use C++ exceptions internally instead of endlessly nesting try-catch expressions against exceptions thrown from the OpenCL C++ wrapper.

Given that the code is already uses proper C++ idioms, it's easier to just use exceptions internally. Instead, the exceptions are now caught at the top level only and converted to status codes to preserve the existing library API.

Note that we deliberately do not catch C++ runtime errors (such as `std::bad_alloc`) nor logic errors (aka failed assertions) in the C++ API because no actual handling can ever happen for such errors.

However, in the C interface wrapper _all_ exceptions (even unknown ones) are caught and converted to the "catch-all" `kUnexpectedError` code.

Related to issue #110.


CNugteren:
First of all thanks for the work!

It is a long list of changes so it took me some time to go through everything. Of course many are just small changes that show up like large changes (e.g. un-indenting a whole function). In general I am quite happy with the changes and I think it improves the library a lot.

There is one thing do that I would like to ask you to change if possible. This concerns the `clpp11.hpp` file. This is part of https://github.com/CNugteren/CLCudaAPI and is thus meant to be independent of CLBlast and single-header only. Right now this changed due to the new exception class. Would it be possible to make the header stand-alone again, keeping in mind that all changes will have to be made (and should be possible to make) in the CUDA version of the CLCudaAPI header? I hope this is not to difficult to achieve. I can also help if you want.

CNugteren:
Indeed, those two options are not so nice. Can we not have a simple exception class in `clpp11.h` that handles all exceptions from OpenCL functions? Or does it really need to inherit from the same base exception class that the other CLBlast exceptions use?

intelfx:
@CNugteren 

I ended up not using inheritance for anything other than just reducing amount of boilerplate code[1]. So, in some sense, it doesn't. However, there are three types of exceptions thrown from there (not just one).

[1]: which is not really nice, but I could not invent a good "semantic" inheritance design. Probably not worth the effort, though.

intelfx:
I could suggest playing a deadly linker dance with weak functions, but that's not portable to MSVC.

CNugteren:
No, let's not do any kind of dances. Perhaps I should just merge it in and see if I can later make this compatible again with the idea of CLCudaAPI and a single interchangeable header for OpenCL / CUDA.

I'll go over the code one more time tomorrow and test it on my machines and then I'll merge it in. Thanks!

intelfx:
Well, probably, I could simplify the exception class hierarchy, leaving only `CLError` used in the CL wrapper header. That way, we'll lose a bit of error code granularity... I'll attempt to do something overnight.

CNugteren:
Yes, perhaps only a `CLError` for all errors in CLCudaAPI's abstraction of OpenCL would make sense.

intelfx:
I mean, the other errors would go into the standard exception classes. This way we'll have to invent a single StatusCode for all non-CL errors reported from the CL header.

intelfx:
...So, I'm afraid, this won't fly (at least strictly as discussed).

In the code I rely on the ability to catch "any runtime error generated by our code" (but exclude e.g. logic errors and C++ runtime errors). This means I have to move all three classes (`Error<Base>`, `ErrorCode<Base, Status>` and `CLError`) into the `clpp11.hpp` header.

While it's technically possible to create a completely separate exception class for CL errors, that would cost us code bloat in several places (in CLBlast). On the other hand, doing as outlined above would mean that we have to replicate the class hierarchy in the second (CUDA) implementation of CLCudaAPI.

I'd rather propose to forgo the "single file header" requirement and move the exception classes into a separate file, say, `cmpp11.hpp` (from "common") that will be common to CL and CUDA implementations of CLCudaAPI.

intelfx:
> Also, I just got the request in #119 to add a prefix for the plain C version of the API. Can this be done easily in the current set-up with a #define or something? If not then perhaps we should split up the error codes again and move them to the 2 respective files. Perhaps it's nice as well to have everything in 1 single header, even if that means code duplication.

Ah, OK.

intelfx:
(Probably) all done, please take a look.

CNugteren:
Thanks again for all your effort! The new version looks good, I've tested in on two system myself as well. I've also added some of the error statuscodes mentioned above.

CNugteren:
Just one more remark. I was testing the new code with MSVC today and noticed the following kind of errors:

```
C:\projects\clblast\src\utilities/clblast_exceptions.hpp(33): warning C4275: non dll-interface class 'clblast::RuntimeError' used as base for dll-interface class 'clblast::ErrorCode<clblast::RuntimeError,clblast::StatusCode>'
c:\projects\clblast\src\cxpp11_common.hpp(52): note: see declaration of 'clblast::RuntimeError'
C:\projects\clblast\src\utilities/clblast_exceptions.hpp(33): note: see declaration of 'clblast::ErrorCode<clblast::RuntimeError,clblast::StatusCode>'
C:\projects\clblast\src\utilities/clblast_exceptions.hpp(33): warning C4251: 'clblast::ErrorCode<clblast::RuntimeError,clblast::StatusCode>::details_': class 'std::basic_string<char,std::char_traits<char>,std::allocator<char>>' needs to have dll-interface to be used by clients of class 'clblast::ErrorCode<clblast::RuntimeError,clblast::StatusCode>'
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE\xstring(2633): note: see declaration of 'std::basic_string<char,std::char_traits<char>,std::allocator<char>>'
```

So is it really necessary to make these exceptions part of the interface of the library? Aren't they all catched or converted to standard C++ exceptions? And if they are: how severe is this warning message and how do we prevent it from happening?

intelfx:
I've simply forgot to remove `PUBLIC_API` attributes from the internally used exception classes (they are a left-over from my first attempt, when I still thought that they will be a public API). Try to remove the attributes from classes in `src/utilities/clblast_exceptions.hpp` and see if errors disappear.

Regarding your API worries: as I have said earlier, I deliberately do not catch "logic error" exceptions, because they cannot be handled in any meaningful way. However, all of our exception classes inherit from the default ones, so (to the best of my knowledge) they need not to be public/exported even if we do throw them across shared object boundaries.

(However, if you want, we can catch them too — I thought that there were no objections to this part.)

CNugteren:
OK, that's clear. I've removed `PUBLIC_API` in two occurrences.

psyhtest:
@intelfx Great work!

intelfx:
@psyhtest This is actually a kind of "groundwork" for the plugin system (for which I'm writing an RFC/discussion issue right now). Considering the issues with linkage to our Mali GEMM overlay library on the devboard (and also keeping in mind the risk of ABI incompatibilities), I now think that overlaying symbols (either via linkage or LD_PRELOAD) isn't really a good approach.

psyhtest:
@intelfx Excellent. Look forward to discussions on the plugin system!

PR #131, opened 2017-01-20
Assorted minor fixes

intelfx:
Minor fixes that materialized during plugins system development (relevant PR will follow shortly). See commit messages.

### DIFF
--- a/test/correctness/tester.cpp
+++ b/test/correctness/tester.cpp
@@ -248,8 +248,29 @@ template <typename T, typename U>
 void Tester<T,U>::TestErrorCodes(const StatusCode clblas_status, const StatusCode clblast_status,
                                  const Arguments<U> &args) {
 
+  // Either an OpenCL or CLBlast internal error occurred, fail the test immediately
+  // NOTE: the OpenCL error codes grow downwards without any declared lower bound, hence the magic
+  // number. The last error code is atm around -70, but -500 is chosen to be on the safe side.
+  if (clblast_status != StatusCode::kSuccess &&
+      (clblast_status > static_cast<StatusCode>(-500) /* matches OpenCL errors (see above) */ ||
+       clblast_status < StatusCode::kNotImplemented) /* matches CLBlast internal errors */) {
+    PrintTestResult(kErrorStatus);
+    ReportError({StatusCode::kSuccess, clblast_status, kStatusError, args});
+    if (verbose_) {
+      fprintf(stdout, "\n");
+      PrintErrorLog({{StatusCode::kSuccess, clblast_status, kStatusError, args}});
+      fprintf(stdout, "   ");
+    }
+  }
+
+  // Routine is not implemented
+  else if (clblast_status == StatusCode::kNotImplemented) {
+    PrintTestResult(kSkippedCompilation);
+    ReportSkipped();
+  }
+
   // Cannot compare error codes against a library other than clBLAS
-  if (compare_cblas_) {
+  else if (compare_cblas_) {
     PrintTestResult(kUnsupportedReference);
     ReportSkipped();
   }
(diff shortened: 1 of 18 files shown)


intelfx:
I need advice on the last commit.

Currently, the error code checking is completely skipped if the reference BLAS is not clBLAS. This leads to funny behavior like missing obvious internal errors, reporting `0 passed / 64 skipped / 0 failed` and hence "passing" the test when half of the errors are actually from clCreateKernel.

That last commit fixes that by always considering OpenCL API and internal CLBlast errors, but OpenCL error codes grow downwards without any known "last" code, so I just wrote `clblast_status > static_cast<StatusCode>(-100)` in the conditional.

Do you have any better ideas?

test/correctness/tester.cpp
```diff
@@ -248,8 +248,22 @@ template <typename T, typename U>
...
+  // Check for OpenCL and internal errors
+  // (their mere presense indicates a test failure)
+  if (clblast_status != StatusCode::kSuccess &&
+      (clblast_status > static_cast<StatusCode>(-100) /* OpenCL API error */ ||
+       clblast_status < StatusCode::kNotImplemented) /* CLBlast internal error */) {
+    PrintTestResult(kErrorStatus);
+    ReportError({StatusCode::kSuccess, clblast_status, kStatusError, args});
+    if (verbose_) {
+      fprintf(stdout, "\n");
+      PrintErrorLog({{clblas_status, clblast_status, kStatusError, args}});
```
CNugteren: This `clblas_status` should probably also be `StatusCode::kSuccess`.

CNugteren:
Thanks for the improvements! Some remarks before I merge this in:

1. `Routine: use PrecisionSupported<>() instead of duplicating the check`: Makes total sense, thanks.
2. `treewide: include clpp11.hpp first to silence deprecation warnings. Otherwise, cl.h gets included through clblast.h before clpp11.hpp. `: That explains a lot, I couldn't get these warnings silenced before when I tried. Now it should be fine indeed.
3. `samples: add CL_USE_DEPRECATED_OPENCL_1_*_APIS where needed`: See point 2, thanks.
4. `src/clpp11.hpp: do not store program source/binary in Program`: If I recall correctly this was added in [CLCudaAPI](https://github.com/CNugteren/CLCudaAPI) for compatibility with the CUDA back-end. But for CLBlast the CUDA back-end is not used, so I guess this change is good, especially since it could help for performance. If it is ever needed again we can bring it back easily I guess.
5. `Routine: fix semi-warm routine construction (when binary is in cache)`: That's a good catch! The return statement was lost in the big exception refactoring changes.
6. `src/clpp11.hpp: check pointers before clRelease*()`: OK, why not indeed.
7. `FillCache: perform compilation for each precision separately, catch e`: Looks good, also cleans it up a bit. It does require 15 extra lines, so I believe the `scripts/generator/generator.py` file needs a small modification at line 45 (second item of `FOOTER_LINES` should be 15 larger).

`Tester: always fail on OpenCL and CLBlast internal errors`: I agree, OpenCL errors and internal errors should be errors instead of 'skipped' tests. This was originally implemented in this way when I was only comparing to clBLAS and when I was not using automated tests. This change makes sense now. About the -100, I think that is fine right now. Or perhaps even -500? Maybe you can document it a little bit more in a comment, e.g. "-70 is the current last OpenCL error code but it grows with new standards, so -500 is chosen to be on the safe side".

Note that the following snippet is now obsolete:
```
  // Could not compile the CLBlast kernel properly
  else if (clblast_status == StatusCode::kOpenCLBuildProgramFailure ||
           clblast_status == StatusCode::kNotImplemented) {
    PrintTestResult(kSkippedCompilation);
    ReportSkipped();
  }
```
Perhaps you can still make it such that `PrintTestResult(kSkippedCompilation)` is printed in such cases but that it is counted as an actual error as in your commit?

test/correctness/tester.cpp
intelfx: Yep, thanks.

intelfx:
@CNugteren: Done.

1. I've retained the `kNotImplemented` conditional and moved it before CBLAS/clBLAS check (not-implemented routines will remain so regardless of reference BLAS); is this OK?

2. Please take a look on the two new commits.

intelfx:
Hm, I've apparently understood what you were talking about... You wanted me to `PrintTestResult(kSkippedCompilation)` instead of `kErrorStatus` if status is `StatusCode::kOpenCLBuildProgramFailure` inside the first (newly added) conditional, right?

It looks like too much bikeshedding (sorry), but I guess that "attempted and failed to compile" is not the same as "skipped compilation", so I'd rather leave it as it is.

And a few changes again — I missed some warnings.

CNugteren:
I agree with your latest changes, it's fine. Most importantly is that it actually reports an error (and thus returns an error code when running e.g. `CTest`) when there is an OpenCL error or when compilation fails. Also thanks for catching the OS X / gcc issue. There is gcc on OS X, but it seems to be a symlink to clang.

I'll merge this in now.

Issue #130, opened 2017-01-11
Passing tuning parameters manually to routines


ari-wwu:
Hello,

is there there a possibility to pass tuning parameter values manually to the routines, e.g. gemm?
For example, in case of the gemm routine, I will determine values for its tuning parameters (MWG, NWG, ...) that are also optimized for the input matrices' sizes (M,N and K). Assumed I have determined such optimized tuning parater values for different matrix sizes, can I pass these to the CLBlast library?

Many thanks in advance.

blueberry:
Just to add that I would Also like if such thing is available through API, but I guess that the compiler would not be able to do some optimizations?

CNugteren:
Indeed, this is something that should be added to CLBlast. I will continue with my current efforts (implementing triangular solvers) and afterwards work on this request. Note that some ground-work is already in place, it is possible to [override the default database](https://github.com/CNugteren/CLBlast/pull/84/commits/e4e1f05079273f60f4f15280b3f103810c7eb31f).

I see two options to implement this:
1. A C++ API to provide database overrides, e.g. `clblast::OverrideParameters("gemm", {{"MWG", 8}, {"NWG":32}}`. This allows fine-grained control. This can be useful if you want to supply fine-tuned parameters, e.g. in one program for a specific matrix-size and then afterwards supply other parameters for another matrix-size. This would require either storing multiple compiled OpenCL kernels in the internal cache (one per parameter setting) or possibly just invalidating the cache if the parameters change.
2.An environmental variable (e.g. `CLBLAST_DATABASE_OVERRIDE`) pointing to a file on disk with tuning parameter overrides. This way the user has less control, but it might have the advantage that such a file is easy to edit and exchange (e.g. JSON) and could be automatically generated by the tuners.

My current preference would be for point 1, but I welcome any opinions or suggestions for alternatives.

CNugteren:
Or perhaps a more natural (and easier to implement?) third option is an optional argument to each CLBlast routine which allows you to specify one or more parameter overrides. This way there is no additional global state required (apart from the existing cache) and it is clear from the user's perspective when parameters are overridden or defaulted.

By the way, are you using the C++ API or do you also need this feature in the C version?

blueberry:
I am using CLBlast from Clojure and JVM, so C Is the only option for me.

CNugteren:
@ari-wwu 
I've started working on this. Currently I have the following interface defined:
```
[...]

PR #132, opened 2017-01-20
Refactor cache subsystem

intelfx:
In this PR I rewrite the cache subsystem almost-from-scratch, making it useful for caching arbitrary objects in different cache instances and using `std::map` internally.

The code gets much cleaner, but there is a downside: **it needs C++14**, in particular [N3465](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2012/n3465.pdf) and [N3657](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3657.htm) to allow searching by a tuple of references. If the concern is VS compatibility, then this feature is supported since VS2015 (MSVC 14.0).

In worst case I can change the code to use manual O(n) heterogeneous lookup in a vector, which won't need C++14.

Note that this is on top of PR #131.

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -20,103 +20,73 @@
 namespace clblast {
 // =================================================================================================
 
-// Stores the compiled binary or IR in the cache
-void StoreBinaryToCache(const std::string &binary, const std::string &device_name,
-                        const Precision &precision, const std::string &routine_name) {
-  #ifdef VERBOSE
-    printf("[DEBUG] Storing binary in cache\n");
-  #endif
-  binary_cache_mutex_.lock();
-  binary_cache_.push_back(BinaryCache{binary, device_name, precision, routine_name});
-  binary_cache_mutex_.unlock();
-}
+template <typename Key, typename Value>
+template <typename U>
+Value Cache<Key, Value>::Get(const U &key, bool *in_cache) const {
+  std::lock_guard<std::mutex> lock(cache_mutex_);
 
-// Stores the compiled program in the cache
-void StoreProgramToCache(const Program &program, const Context &context,
-                         const Precision &precision, const std::string &routine_name) {
-  #ifdef VERBOSE
-    printf("[DEBUG] Storing program in cache\n");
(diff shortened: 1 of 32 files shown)


CNugteren:
Again thanks a lot! Let's merge the other PR in first and then I'll review this one. But the idea makes sense, the code becomes cleaner for each routine indeed.

I like C++14. However, I don't think that's a good idea at this point. In fact, I'm not even using all of C++11, I try to make sure it compiles with GCC 4.7 and VS 2013. Even GCC 4.7 might save a lot of people a lot of hassle in Linux, not having to compile it from source or installing it from an unofficial repository. So I would like to keep this if it is not crucial.

One option would be to have a #ifdef for the C++14 part and a less pretty alternative otherwise. What do you think?

intelfx:
OK, I'll change it back to use a vector with hand-rolled O(n) lookup.

Well, it was worth trying :P

intelfx:
#ifdef'ing the C++14 part with a fallback would be nice, but that needs additional logic on CMake side to pick the most recent available C++ standard (which could've been easier if we had 3.1, but I suppose we're stuck with 2.8 due to the same reasons).

intelfx:
Well, actually, in a layman's cache "benchmark" the performance differences are minimal (not even saying that in real world all this would be dominated by GPU time by orders of magnitude). So for now I won't even bother implementing a C++11/C++14 detection in CMakeLists.txt... (except in the case you want me to.)

intelfx:
Rebased on top of `development`.

CNugteren:
Thanks for the additions! I've looked through the code and it looks fine. It is definitely easier to add other caches now. The changes might also come in handy for #130. But I guess then we'll have to tweak it a bit: make it possible to remove or replace an item in the cache.

I'll merge it in now and do some tests myself as well.

intelfx:
Well, implementing cache entry removal is a matter of ~10 extra lines, and replacing a cache entry is also possible right now, because all objects that we cache are refcounted and can be modified in-place.

PR #135, opened 2017-02-03
Plugin framework

intelfx:
This patch series adds support for dynamically loadable plugins that can override host routines, OpenCL kernels and database entries on a per-device/routine/precision basis. Specifically, this is a better and broader replacement for previous "database overlay" code which relies on replacing symbols using LD_PRELOAD (which is implementation-defined at best and UB at worst, and somehow failed to work on one of our test devices).

The plugins are just platform-specific dynamic libraries. For now I only wrote the POSIX backend; Win32 has corresponding functionality in form of `LoadLibrary()`, `GetProcAddress()` and friends (if I recall correctly) which can also be implemented if/when you accept the general approach.

The Plugin<->CLBlast API/ABI is built on top of C++ RTTI and virtual functions; ABI compatibility is ensured via RTTI. All relevant types are derived from a never-changed empty base class and put into an inline namespace whose name is to be changed every time the API or ABI is changed. Hence, by dynamic_casting such a base class pointer one can ensure that a plugin was built against the same definitions as CLBlast.

Host routine override is implemented without any clever tricks, using routine-specific glue code in each host routine. As this is 0-th iteration, I have only modified the GEMM routine (because actually for now we only care about overriding GEMM). If/when this is going to be accepted, I can add glue code to all remaining routines (or, if you wish, just leave it out and do as needed).

Please review this thoroughly, I could easily have made many style/correctness issues. ;)


CNugteren:
OK, lot's of changes. I'll take a look at it in the coming days.

By the way, Windows builds now fail it seems it.

intelfx:
The only way for me to test Windows builds is to actually make a pull request :)

Should be fixed now.

CMakeLists.txt
```diff
@@ -145,6 +145,23 @@ if(CLIENTS OR TESTS)
...
+  )
+endif()
+
+if(NOT PLUGINS)
+  message(STATUS "Disabling plugin support")
+  set(SOURCES_PLUGINS
+    src/utilities/plugin_none.cpp
+  )
+endif()
+
```
CNugteren: Just an idea: would it be useful to guard all the Plugin-related things with a CMake option (e.g. similar to the current SAMPLES or CLIENTS options)? Not sure if the remainder of the code is now dependent on the plugin-system, but if it isn't, this might be a nice small addition. Most users won't need the plugin-system I guess, and this way it might be easier to support older compilers. It also saves another installed header.

include/clblast_plugin.h
```diff
@@ -44,6 +44,47 @@ inline namespace version_1 {
...
+  std::string kernel;
+
+  // Input data precision corresponding to these parameters
+  Precision precision;
+
+  // Entries for this routine
+  std::vector<Vendor> vendors;
+};
+
+} // namespace database
```
CNugteren: All of the above is already elsewhere (`database.hpp`). I understand that it now becomes part of the interface, but perhaps the database code can use these new definitions instead? Problem would be the plugin namespace I guess. If not possible then it's no big deal I guess. Perhaps good to add a pointer to both files to the other file in case one of these structs ever gets changed.

src/database/database.cpp
```diff
@@ -64,7 +65,7 @@ const std::unordered_map<std::string, std::string> Database::kVendorNames{
 // Constructor, computing device properties and populating the parameter-vector from the database.
 // This takes an optional overlay database in case of custom tuning or custom kernels.
```
CNugteren: Minor note: the comment is still referring to the old overlay database, not to the new plugin system

include/clblast_plugin.h
CNugteren: Nevermind my comment, I should've looked at the other changes first, this is exactly what you've done 👍

include/clblast_plugin.h
```diff
@@ -101,12 +101,28 @@ class Interface : public Base {
...
+  enum class KernelMode {
+    // Custom kernel is ignored
+    Default = 0,
+
+    // Custom kernel is appended to built-in PadCopyTransform code
+    NeedPCT,
+
+    // Custom kernel replaces all built-in code
+    Custom
+  } kernel_mode;
```
CNugteren: Maybe this can be done a bit clearer with less whitespace? I would also add integers to the other enums, at first for me it wasn't clear what these things are, e.g:
```
enum class KernelMode {
   Default = 0, // Custom kernel is ignored
   NeedPCT = 1, // Custom kernel is appended to built-in PadCopyTransform code
   Custom = 2 // Custom kernel replaces all built-in code
 } kernel_mode;
```
Another option is to use the [Hungarian notation](http://stackoverflow.com/questions/1230854/objective-c-why-do-constants-start-with-k), similar to the other enums in CLBlast?

CMakeLists.txt
intelfx: Well, `Database` inherits from `plugin::Database`, and the whole idea is to use a stub plugin entry (that does not override anything) if no actual plugin is found instead of cluttering glue code with null pointer checks. So, unless we #ifdef out everything plugin-related, we cannot skip compilation of plugin-related sources (though we can avoid installing the extra header, of course).

include/clblast_plugin.h
CNugteren: Also, in another commit you mention that you don't need the `NeedPCT` option anymore? If so, the routine.cpp double-if becomes a simpler if-else.

src/database/database.cpp
intelfx: Thanks, will fix in next iteration.

include/clblast_plugin.h
intelfx: OK, will do. That's why I have said "please review [code style] thoroughly" :)

src/routine.cpp
```diff
@@ -24,6 +24,7 @@ namespace clblast {
 // The constructor does all heavy work, errors are returned as exceptions
 Routine::Routine(Queue &queue, EventPointer event, const std::string &name,
                  const std::vector<std::string> &routines, const Precision precision,
+                 std::initializer_list<const char *> source_pct,
```
CNugteren: What's the full name of `source_pct`? At first I read 'ptr', perhaps it is a good to idea to use to full name (if not too long of course)

src/routines/level3/xgemm.cpp
```diff
@@ -33,10 +33,11 @@ Xgemm<T>::Xgemm(Queue &queue, EventPointer event, const std::string &name):
     #include "../../kernels/level3/convert_symmetric.opencl"
     #include "../../kernels/level3/convert_triangular.opencl"
     #include "../../kernels/level3/convert_hermitian.opencl"
+    }, {
+    // keep in mind C1091 in MSVC 2013
```
CNugteren: Still good maybe to separate the following 2 sets of 3 includes for MSVC 2013. They contain most of the code anyway, so the issue might still be there, even if you removed some parts. Or have you tested it?

src/routine.cpp
intelfx: pad/copy/transpose — this contains the code for transformation routines (that can be used with custom main kernel).

CNugteren: Ah, PadCopyTranspose :-) Perhaps this is no longer needed then if you don't want to treat that separately?

src/routines/level3/xgemm.cpp
intelfx: Isn't this tested in AppVeyor? Or did they lift that limitation in newer versions of MSVC++?

include/clblast_plugin.h
CNugteren: I guess this also means that the pad-copy-transpose code can be removed from `src/clblast_plugin.cpp`?

src/routine.cpp
intelfx: Why I don't want to treat this separately?

src/routines/level3/xgemm.cpp
CNugteren: No, AppVeyor is MSVC 2015 (at least how I configured it, not sure if I can also use an older compiler - haven't tried it). I have a specific machine in a corner at home with MSVC 2013, I occasionally test it.

src/routine.cpp
intelfx: This is done to allow using built-in pad/copy/transpose code together with a custom main kernel, calling the now-exported `PadCopyTransposeMatrix<T>()` from the custom host code. Unless there is a misunderstanding, I *do* want to have this split.

include/clblast_plugin.h
intelfx: Why?

CNugteren:
In general, again, thanks a lot for the changes. Most of this code I couldn't have (easily) written, so it much appreciated. Not sure if it was the easiest path to get better performance on ARM Mali, but it certainly is flexible and re-usable for other (currently unknown) issues.

I have gone through all the code now (some comments inlined), but I trust you on the general design of the plugin system. In the coming days I will compile it on various systems and run some tests to be sure everything still works.

What's still missing perhaps is some kind of documentation for users, perhaps a small sample plugin program (if small is possible at all) and/or a section in the README. Also an entry in the changelog is nice to have, but I can also add that later.

CMakeLists.txt
CNugteren: Hmm, yes, #ifdef's would not be nice. I made this remark a bit too early perhaps, before I saw all the changes. If it's not easy then let's not bother.

src/routine.cpp
CNugteren: OK, then I misunderstood your commit message:
```
Initially I have intended to have a third routine mode
(RoutineMode::NeedPCT) that will run pad/copy/transpose using the
built-in host code according to layout requirements specified in other
flags, and then run plugin-provided host code for the main kernel, but
for now all custom kernels that I considered require completely
different code on host side, so this feature won't be used.
```

intelfx:
> Not sure if it was the easiest path to get better performance on ARM Mali

Well, almost certainly making a device-specific override is way easier than inventing a new generalized kernel that gives the same performance on a broad set of hardware...

And I still don't quite understand what did you mean by removing the pad-copy-transpose code from `clblast_plugin.cpp`.

include/clblast_plugin.h
intelfx: Oh. I've replied to the first comment of yours (the one where you were saying about style issues); I didn't see the second one. Where did I say that I don't need `NeedPCT` anymore?

src/routine.cpp
intelfx: I meant the third routine mode, not kernel mode. There are two enums.

Initially I wanted to provide a way for a plugin to override only a part of the host routine code (i. e. the built-in routine code will call built-in PadCopyTranspose, and then the custom routine code will call the custom main kernel), but this is too inflexible.

include/clblast_plugin.h
intelfx: Got it, nevermind. GitHub doesn't properly notify me about new replies in inline comments...

src/routine.cpp
intelfx: So now we have two routine modes (either the built-in host code or the custom host code) and three kernel modes (either the built-in OpenCL kernel, or the built-in pad/copy/transpose routines + custom main kernel, or fully custom kernel).

intelfx:
> What's still missing perhaps is some kind of documentation for users, perhaps a small sample plugin program (if small is possible at all) and/or a section in the README. Also an entry in the changelog is nice to have, but I can also add that later.

Sure, I'll add that after we settle on the design and the code.

psyhtest:
@intelfx Nice work!

src/routine.cpp
CNugteren: OK, I understand your remark now. So then nevermind some of the comments I made. Indeed, GitHub can be a tricky with showing the comments, especially this one since it is hidden in history of an outdated change.

CNugteren:
> Well, almost certainly making a device-specific override is way easier than inventing a new generalized kernel that gives the same performance on a broad set of hardware... I'd rather say that plugins are the quick'n'dirty way, as opposed to inventing a better generalization.

In that sense, yes. But we could've also just added the ARM specific code into the library and only run it if we are on an ARM platform. That's what I meant. Anyway, doesn't matter.

> And I still don't quite understand what did you mean by removing the pad-copy-transpose code from clblast_plugin.cpp.

I believe you found my remark now, it is hidden in the outdated `src/routine.cpp` changes. 

To make sure we are on the same page, here is what I understand. There are 2 routine modes now:

* Running the original host code
* Running custom host code: replacing the whole function in one go, not reusing anything

And 3 kernel modes:

* Running the original kernels
* Running fully custom kernels: replacing all the kernel code, not reusing anything
* Running a custom main kernel but reusing pad/copy/transpose kernels

And it is possible to mix any of the above, or are there restrictions? So 6 modes in total if you combine them? Please correct me if I'm wrong. For the last mode, I understand that you would want to split the current library's included kernel code in two parts. However, I don't understand why the plugin system would need access to the host code for the pad/copy/transpose kernels.

intelfx:
> To make sure we are on the same page, here is what I understand. There are 2 routine modes now <...> and 3 kernel modes <...>

Yes, correct.

> And it is possible to mix any of the above, or are there restrictions? So 6 modes in total if you combine them?

Yes, there are no artificial restrictions. Technically, each combination is possible. However, some of them do not make much sense: the built-in OpenCL kernels are not part of the interface, so a plugin that uses the built-in host code with a custom OpenCL kernel (or vice versa) will only work until you decide to change the order of arguments for some reason.

I should probably document that for the next iteration.

> For the last mode, I understand that you would want to split the current library's included kernel code in two parts. However, I don't understand why the plugin system would need access to the host code for the pad/copy/transpose kernels.

For the same reason: the host p/c/t code is now an interface, while the OpenCL kernels are not. This way, we can make changes to the kernels and compensate in the host wrappers to keep the interfaces stable, should such need arise in the future.

CNugteren:
OK, I understand. So the alternative to exposing the pad/copy/transpose function in the plugin interface would be to write custom kernels only, but then there is no guarantee that the interface to the kernel doesn't change. Not super-nice indeed because there is no way to check from the host C++ code, but still the current approach also has its disadvantages it seems:

1) There is quite some extra code needed just for the pad/copy/transpose function. More than for supporting GEMM it seems.
2) The pad/copy/transpose routine is now considered a special function, perhaps others will need to be added as needed in the future.
3) The user will have to re-write the pad/copy/transpose routine if he wants to change the kernel just to make sure the interface will not change in the future. Do you think all users will take that extra effort?

I'm not saying that we should change this necessary, I'm mostly just having a discussion to make sure I understand everything properly :-)

intelfx:
> So the alternative to exposing the pad/copy/transpose function in the plugin interface would be to write custom kernels only, but then there is no guarantee that the interface to the kernel doesn't change.

Hm, I don't quite understand this sentence... If we decide not to expose P/C/T host code, then we must either disallow using built-in P/C/T kernels together with custom main kernels, or make the P/C/T kernels part of the stable interface.

> There is quite some extra code needed just for the pad/copy/transpose function. More than for supporting GEMM it seems.

Which extra code do you mean — the `PadCopyTransposeMatrix<T>()` wrapper or the corresponding conditional branches in `Routine::Routine()`? There isn't much code IIUC, most of it are mechanical instantiations...

> The pad/copy/transpose routine is now considered a special function, perhaps others will need to be added as needed in the future.

Yep. Something must be made an interface. Again, either we stabilize the host routine interface, or we stabilize the OpenCL kernel interface, or both, or forbid such "code reuse" in plugins at all.

> The user will have to re-write the pad/copy/transpose routine if he wants to change the kernel just to make sure the interface will not change in the future. Do you think all users will take that extra effort?

Those who provide custom P/C/T kernels will always have to provide custom host P/C/T code, unless we stabilize _both_ kernel interface AND the host code interface. This can be done in the future, should the need arise.

CNugteren:
Let me give an example of an alternative way of supplying a custom Pad/Copy/Transpose or whatever kernel to CLBlast. Let's say we want to override the `CopyFast` kernel. We could guard the original kernel like this in the library:
```
#ifndef CUSTOM_COPY_MATRIX_FAST
void CopyMatrixFast(const int ld,
                    __global const realC* restrict src,
                    __global realC* dest,
                    const real_arg arg_alpha) {
   (...) // current CLBlast implementation here
}
#endif
```

The supplier of the custom kernel can then just write a kernel-string and pass it to CLBlast, as long as it contains `#define CUSTOM_COPY_MATRIX_FAST` and is included before the original source code (but perhaps after the `common.opencl` header?), e.g.:
```
#define CUSTOM_COPY_MATRIX_FAST
void CopyMatrixFast(const int ld,
                    __global const realC* restrict src,
                    __global realC* dest,
                    const real_arg arg_alpha) {
   (...) // custom implementation here
}
```
This can even be done as part of the normal API, no complicated plugin system needed. Of course this way you're restricted to the launch configuration of the original kernel.

To be clear: I see the value of the plugin system for host code, but not so much for a custom kernel only. That's why it's still not clear to me why the Pad/Copy/Transpose code needs to be treated separately.

intelfx:
> This can even be done as part of the normal API, no complicated plugin system needed.

One of the main values of this plugin system is that it's transparent to the calling application: the integrator just puts a dynamic library somewhere in the library search path, and the rest happens automatically. If we change the API, then it's not so anymore.

> To be clear: I see the value of the plugin system for host code, but not so much for a custom kernel only.

(see above)

> The supplier of the custom kernel can then just write a kernel-string and pass it to CLBlast, as long as it contains #define CUSTOM_COPY_MATRIX_FAST and is included before the original source code

...and this is a very nice idea, thanks. It raises one question, though: is there any penalty for including unused functions in the kernel?

CNugteren:
> One of the main values of this plugin system is that it's transparent to the calling application: the integrator just puts a dynamic library somewhere in the library search path, and the rest happens automatically. If we change the API, then it's not so anymore.

Yes, that's true of course. Another option for loading a custom kernel source would be to point an environment variable to a specific location on disk and then load that file. That has its drawbacks as well. And you need the plugin system also for the host code, so it is already there of course.

> ...and this is a very nice idea, thanks. It raises one question, though: is there any penalty for including unused functions in the kernel?

What do you mean exactly? Unused functions have to be compiled of course, so that takes time. But if they are guarded with #ifndef then it's just a string that gets a bit larger, that should be no problem at all.

So what do you think about using the plugin system only for overriding the main routine functions and then not exposing the Pad/Copy/Transpose function? Of course this has the issue that the kernel interface might change over time and this is not nicely handled at source level, but I don't think that's too bad: a specific plugin might be bound to a specific CLBlast version.

intelfx:
> What do you mean exactly? Unused functions have to be compiled of course, so that takes time. But if they are guarded with #ifndef then it's just a string that gets a bit larger, that should be no problem at all.

Let's imagine a custom kernel and host routine that do not require input transformations at all (like the "direct" version of the existing GEMM kernel). Yet the P/C/T functions are nevertheless compiled and sent to the device. Does this incur any run-time performance penalties on the device?

If "no", then indeed we can just guard the built-in kernels with #ifndefs and include them after the custom kernel string.

> So what do you think about using the plugin system only for overriding the main routine functions and then not exposing the Pad/Copy/Transpose function?

Uh? I'm afraid I object to not exposing `PadCopyTransposeMatrix<T>()`, because I don't want to reimplement it in my plugin (or copy-paste it together with clpp11.h, for that matter).

CNugteren:
> Let's imagine a custom kernel and host routine that do not require input transformations at all (like the "direct" version of the existing GEMM kernel). Yet the P/C/T functions are nevertheless compiled and sent to the device. Does this incur any run-time performance penalties on the device?
 
Not sure if the whole binary is actually sent to the device or whether it works per function. But I think in either case such performance penalties will only be of a concern for very small computations. Perhaps someone with more knowledge of an OpenCL compiler and run-time (@psyhtest?) can give a proper answer.

> If "no", then indeed we can just guard the built-in kernels with #ifndefs and include them after the custom kernel string.

With the #ifndefs you can actually even disable the kernels that you don't need at all (e.g. if your custom GEMM is a 'direct' GEMM and doesn't need any copy/pad/transpose). Right?


> Uh? I'm afraid I object to not exposing PadCopyTransposeMatrix<T>(), because I don't want to reimplement it in my plugin (or copy-paste it together with clpp11.h, for that matter).

Then I still don't understand your earlier remark it seems. There is some miscommunication, perhaps I should re-read some of your comments from the start. To me what you said earlier seems in conflict with what you just said above:

> Initially I have intended to have a third routine mode
(RoutineMode::NeedPCT) that will run pad/copy/transpose using the
built-in host code according to layout requirements specified in other
flags, and then run plugin-provided host code for the main kernel, but
for now all custom kernels that I considered require completely
different code on host side, so this feature won't be used.

and:

> Initially I wanted to provide a way for a plugin to override only a part of the host routine code (i. e. the built-in routine code will call built-in PadCopyTranspose, and then the custom routine code will call the custom main kernel), but this is too inflexible.

So for my understanding you have two use-cases:
* You want to provide custom kernels only. This you can do with just providing an extra argument to the Routine base-class.
* You want to override the whole host and kernel code. In this case (as you say in the remark above) you'll re-write everything from scratch because it will be custom anyways. This is what I understand based on "all custom kernels that I considered require completely different code on host side, so this feature won't be used".

intelfx:
> With the #ifndefs you can actually even disable the kernels that you don't need at all (e.g. if your custom GEMM is a 'direct' GEMM and doesn't need any copy/pad/transpose). Right?

Ah, yes. My bad. One can just #define these flags and not provide replacement kernels at all. Then this is not an issue and we probably should go this way instead of having `KernelMode` at all.

> You want to provide custom kernels only. This you can do with just providing an extra argument to the Routine base-class.

This is not exactly my usecase, but it is useful for debugging. Yes, as we've just found out, parts of the built-in kernel can be disabled via the preprocessor, so we don't need any knobs for this at plugin API level.

> You want to override the whole host and kernel code. In this case (as you say in the remark above) you'll re-write everything from scratch because it will be custom anyways. This is what I understand based on "all custom kernels that I considered require completely different code on host side, so this feature won't be used".

Yes. In this mode the host code would probably be rewritten from scratch. However, it may still want to pad/transpose input matrices using generic kernels. This is the separation of "policy" (how to pick paddings, etc.) and "mechanism" (how to call the pad/transpose kernel).

CNugteren:
I'm glad it seems the discussion was not for nothing 🎉. Sometimes it's so much easier to understand each other face-to-face compared to writing comments on a PR.

> Ah, yes. My bad. One can just #define these flags and not provide replacement kernels at all. Then this is not an issue and we probably should go this way instead of having KernelMode at all.

OK, that seems like a better solution to me indeed. It might be useful for other scenario's as well later on. And then we also don't have to treat the pad/copy/transpose kernels separately. Will you add the guards for all kernels in this PR?

> Yes. In this mode the host code would probably be rewritten from scratch. However, it may still want to pad/transpose input matrices using generic kernels. This is the separation of "policy" (how to pick paddings, etc.) and "mechanism" (how to call the pad/transpose kernel).

I understand. Indeed, some utility functions can be nice to have access to, including the pad/copy/transpose host function. There is already the Xomatcopy routine, but that doesn't do the padding I believe, so we'll need to expose the pad/copy/transpose routine separately indeed. Other utility functions are probably small enough not to have to bother with - they can be re-written by the plugin developer? Or the plugin developer can even decide to copy-paste some CLBlast code into the plugin library?

intelfx:
> Will you add the guards for all kernels in this PR?

Yep. Next iteration. I'm currently a little busy with the university, so expect that in Monday.

> There is already the Xomatcopy routine, but that doesn't do the padding I believe

...and no event wait list, another Routine context... calling another CLBlast routine to do pad/transpose sounds like too high-level.

> Other utility functions are probably small enough not to have to bother with - they can be re-written by the plugin developer? Or the plugin developer can even decide to copy-paste some CLBlast code into the plugin library?

Which exactly? I suppose, if the need arises, we can always export more helpers. But yes, one-liners are not worth it.

CNugteren:
> Yep. Next iteration. I'm currently a little busy with the university, so expect that in Monday.

OK, no hurries of course!

> Which exactly? I suppose, if the need arises, we can always export more helpers. But yes, one-liners are not worth it.

Not sure, perhaps the valid-buffer-testers? But it might change in the future. Anyway, doesn't matter now.

CNugteren:
By the way, you might also want to have a quick look at the changes I made to the database in the `custom_parameters` branch. Basically I split the big database into multiple smaller databases. Thanks to your efforts each of those smaller databases can be cached. I did this because of #130: I can now remove a per-kernel database from the cache and replace it with a user-custom version for a specific kernel.

psyhtest:
@CNugteren It's a valuable contribution, Unfortunately, @intelfx has no time to finish it off. Would you perhaps be able to do so yourself?

CNugteren:
@psyhtest Yes, I should. I have been a bit busy the last few weeks. My plan for now is first to get a 1.0 release of CLBlast. That means solving open (bug/performance) issues and fixing all automated tests at http://ci.arrayfire.org/view/Other/. Afterwards I will continue this PR.

psyhtest:
> My plan for now is first to get a 1.0 release of CLBlast out.

Go for it! :)

CNugteren:
As discussed with @psyhtest, I will be taking some time to coming days to revive this pull request. The first step was made today: pulling in all changes from @intelfx into my own branch `plugin_framework` and integrating the latest changes from CLBlast. Next steps:

1. Check whether the merge went OK and everything still compiles
2. Write a unittest or a sample plugin program to test the code
3. Finish the PR by making the changes we discussed

intelfx:
@CNugteren, @psyhtest,

I'm sorry for turning one day into more than three months. Now that I'm done with university... Have you already begin working on finishing this PR or I still have a chance to complete the work?

CNugteren:
@intelfx Sure, your help is very welcome! You should probably be able to do a better job than I at finishing this PR. The only thing I did so far was spend some time merging in the latest changes from `master` (quite some things changed) and at the same time removing some changes that were no longer necessary given our discussion earlier. I suggest that you continue from that point on if you want, let's not do double-work. Perhaps you can branch of of my `plugin_framework` branch (which might not necessarily compile)?

intelfx:
I've already did the rebase and made the changes we've discussed; now I'm writing the documentation, and a sample plugin is TBD.

CNugteren:
Wow, that's fast :+1: 

Feel free to open a new PR once you're ready and we'll see how to proceed.

intelfx:
Why new? Can I force-push here?

CNugteren:
Oh, yeah, sure, as you think is best.

psyhtest:
@intelfx What are you planning to include as a sample? You used Mali-optimised SGEMM in the past, but there's an [updated kernel](https://github.com/ARM-software/ComputeLibrary/blob/master/src/core/CL/cl_kernels/gemm.cl) now available in the ARM Compute Library.

CNugteren:
I would actually try to make the sample as simple as possible, it doesn't necessarily need to lead to a performance improvement on a particular device. It should first of all explain how to use the new plugin interface of course!

It would also be nice to have simple tests to at least see if everything compiles properly and if a simple kernel can be replaced. But I can also write those tests, it is not a big deal.

CNugteren:
@intelfx Any updates? Could you push your changes so I can review them?

CNugteren:
I'm closing this due to inactivity. Also, not sure how relevant this still is. If you think it is, please re-open a new PR based on the latest master of CLBlast. Thanks anyways for your efforts :-)

PR #156, opened 2017-05-24
changing "wb" to "w" when saving json file (text mode)

gfursin:
Hi Cedric,
I encountered a problem when using Anaconda Python 3 on Windows when recording json files. I had to change "wb" to "w" in 2 files when saving json file (text mode) - it seems to be working with both Python 2 and Python 3. Hope it's correct and of any help ;) ...
Thanks,
Grigori


CNugteren:
I believe [you are right](https://stackoverflow.com/a/12309296/4807429)! To be sure I'll test it on some of my set-ups as well and if there are no issues I'll merge it in. Thanks!

gfursin:
Sure, Cedric, thanks!

PR #173, opened 2017-07-17
Add PSO parameters support and search strategy selection from command…

mcian:
Hi Cedric,
 I added the support for PSO parameters and search stategy selection by command line.

We tested the PSO strategy on several architectures and the proposed parameters seem to be effective on finding  good configurations.
Best
Marco

### DIFF
--- a/src/tuning/kernels/xgemm.cpp
+++ b/src/tuning/kernels/xgemm.cpp
@@ -30,7 +30,7 @@ class TuneXgemm {
 
   // The representative kernel and the source code
   // static std::string KernelFamily() { return (V==1) ? "xgemm_1" : "xgemm_2"; }
-  static std::string KernelFamily() { switch(V){ case 1: return "xgemm_1"; case 2: case 3:return "xgemm_2";  }}
+  static std::string KernelFamily() { switch(V){ case 1: return "xgemm_1"; case 2: return "xgemm_2";  }}
   static std::string KernelName() { return "Xgemm"; }
   static std::string GetSources() {
     return
@@ -44,7 +44,7 @@ class TuneXgemm {
   // The list of arguments relevant for this routine
   static std::vector<std::string> GetOptions() {
     return {kArgM, kArgN, kArgK, kArgAlpha, kArgBeta, kArgFraction,
-     kArgHeuristicSelection, kArgMultiSearchStrategy, kArgPsoSwarmSize, 
+     kArgHeuristicSelection, kArgPsoSwarmSize, 
      kArgPsoInfGlobal, kArgPsoInfLocal, kArgPsoInfRandom};
   }
 
@@ -58,12 +58,11 @@ class TuneXgemm {
   static size_t DefaultBatchCount() { return 1; } // N/A for this kernel
   static double DefaultFraction() { return (V==1) ? 1.0 : 512.0; } // test all or sample randomly
   static size_t DefaultNumRuns() { return 2; } // run every kernel this many times for averaging
-  static size_t DefaultNumSearchStragegy() { return 2; } // Full search and Random/PSO
   static size_t DefaultSwarmSizePSO() { return 8; } 
   static double DefaultInfluenceGlobalPSO(){ return 0.1; }
   static double DefaultInfluenceLocalPSO(){ return 0.3; }
   static double DefaultInfluenceRandomPSO(){ return 0.6; }
-  static size_t DefaultHeuristic(){ return (size_t)cltune::SearchMethod::PSO;} // PSO
+  static size_t DefaultHeuristic(){ return static_cast<size_t> (cltune::SearchMethod::PSO);}
   static double DefaultMaxTempAnn(){ return 1.0;}
   
   // Describes how to obtain the sizes of the buffers
@@ -180,29 +179,18 @@ class TuneXgemm {
     return 2 * args.m * args.n * args.k;
   }
   static std::string PerformanceUnit() { return "GFLOPS"; }
-  
+ 
   // Returns which Heuristic to run 
-  static size_t GetCurrentHeuristic(const Arguments<T> &args){
-
-     //  Multi Search Strategy is enable
-    if( args.multi_search_strategy){
-      if( V == 1){ 
-        return (size_t) cltune::SearchMethod::FullSearch;
-      }
-      else{
-        return args.heuristic_selection;
-      }
-    }
-
+  static size_t GetHeuristic(const Arguments<T> &args){
     // Use full-search to explore all parameter combinations or random-search to search only a part of
     // the parameter values. The fraction is set as a command-line argument.
     if (args.fraction == 1.0 || args.fraction == 0.0) {
-      return (size_t) cltune::SearchMethod::FullSearch; 
+      return static_cast<size_t> (cltune::SearchMethod::FullSearch);
     }
     else {
       return args.heuristic_selection;
     }
-  }
+  } 
 };
 
 // =================================================================================================
@@ -223,35 +211,6 @@ void StartVariation(int argc, char *argv[]) {
     case clblast::Precision::kDouble: clblast::Tuner<clblast::TuneXgemm<double,V>, double>(argc, argv); break;
     case clblast::Precision::kComplexSingle: clblast::Tuner<clblast::TuneXgemm<float2,V>, float2>(argc, argv); break;
     case clblast::Precision::kComplexDouble: clblast::Tuner<clblast::TuneXgemm<double2,V>, double2>(argc, argv); break;
-  }
-}
-
-// Test multiple heuristics if kArgMultiSearchStrategy is enabled
-// Otherwise, run the heuristic specified in kArgMultiSearchStrategy
-void TestHeuristic(int argc, char *argv[]){
-
-  auto command_line_args = clblast::RetrieveCommandLineArguments(argc, argv);
-  auto help = std::string{""};
-  auto heuristic_selected = clblast::GetArgument(command_line_args, help, clblast::kArgMultiSearchStrategy, 0); 
-  auto multi_search_strategy = clblast::GetArgument(command_line_args, help, clblast::kArgMultiSearchStrategy, 0);
-
-  if(multi_search_strategy){
-    StartVariation<1>(argc, argv);
-    StartVariation<2>(argc, argv);
-  }
-  else
-  { 
-    switch(heuristic_selected){
-      case 1:
-      case 2:
-      case 3:
-        StartVariation<2>(argc, argv);
-        break;
-      case 0:
-      default:
-        StartVariation<1>(argc, argv);
-        break;
-    }
   }
 }
 
--- a/src/tuning/kernels/xgemm.cpp
+++ b/src/tuning/kernels/xgemm.cpp
@@ -18,6 +18,9 @@
 
 #include "utilities/utilities.hpp"
 #include "tuning/tuning.hpp"
+#define XGEMM_EXEC
+bool tStrategyFlag=true;
+#define DEFAULT_NUM_OF_STRATEGY 1
 
 namespace clblast {
 // =================================================================================================
@@ -28,7 +31,8 @@ class TuneXgemm {
  public:
 
   // The representative kernel and the source code
-  static std::string KernelFamily() { return (V==1) ? "xgemm_1" : "xgemm_2"; }
+  // static std::string KernelFamily() { return (V==1) ? "xgemm_1" : "xgemm_2"; }
+  static std::string KernelFamily() { switch(V){ case 1: return "xgemm_1"; case 2: case 3:return "xgemm_2";  }}
   static std::string KernelName() { return "Xgemm"; }
   static std::string GetSources() {
     return
@@ -41,7 +45,7 @@ class TuneXgemm {
 
   // The list of arguments relevant for this routine
   static std::vector<std::string> GetOptions() {
-    return {kArgM, kArgN, kArgK, kArgAlpha, kArgBeta, kArgFraction};
+    return {kArgM, kArgN, kArgK, kArgAlpha, kArgBeta, kArgFraction,tStrategy, psoSwarmSize, psoInfG, psoInfL, psoInfR};
   }
 
   // Tests for valid arguments
@@ -82,6 +86,7 @@ class TuneXgemm {
       tuner.AddParameter(id, "SB", {0, 1});
     } // a lot more tuning parameters - has to be sampled randomly, too much to test all
     else {
+      //RANDOM_SEARCH & PSO
       tuner.AddParameter(id, "MWG", {16, 32, 64, 128});
       tuner.AddParameter(id, "NWG", {16, 32, 64, 128});
       tuner.AddParameter(id, "KWG", {16, 32});
@@ -97,6 +102,7 @@ class TuneXgemm {
       tuner.AddParameter(id, "SA", {0, 1});
       tuner.AddParameter(id, "SB", {0, 1});
     }
+
   }
 
   // Sets the constraints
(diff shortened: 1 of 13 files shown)


src/utilities/utilities.hpp
```diff
@@ -172,13 +193,13 @@ struct Arguments {
...
-  std::vector<T> alphas; // = {ConstantOne<T>()};
-  std::vector<T> betas; // = {ConstantOne<T>()};
+  std::vector<size_t> x_offsets = {0};
+  std::vector<size_t> y_offsets = {0};
+  std::vector<size_t> a_offsets = {0};
+  std::vector<size_t> b_offsets = {0};
+  std::vector<size_t> c_offsets = {0};
+  std::vector<T> alphas = {ConstantOne<T>()};
+  std::vector<T> betas = {ConstantOne<T>()};
   // Sizes
```
CNugteren: Did you have to change this back? This was actually changed to [fix a compilation issue under MSVC 2013](https://github.com/mcian/CLBlast/commit/f7a16d427c11311e80ae5e3a67bc42b6170f6532). So this will have to go back the way it was unfortunately.

CNugteren:
OK, thanks! I will review the code itself a bit now and we'll discuss more higher-level things on Skype.

src/utilities/utilities.hpp
```diff
@@ -28,6 +28,21 @@
...
+#define RANDOM_SEARCH_STRATEGY 1
+#define PSO_STRATEGY 2
+#define DVDT_STRATEGY 3
+
+#define DEFAULT_STRATEGY 0
+
+#define DEFAULT_PSO_SWARM 8 
+#define DEFAULT_PSO_G 0.3
+#define DEFAULT_PSO_L 0.6
+#define DEFAULT_PSO_R 0.1
```
CNugteren: Could you try to avoid using pre-processor defines? With C++ and especially since C++11 it shouldn't be necessary to use these unfriendly things anymore ;-) You should probably move them to wherever they belong (e.g. in the tuner) and declare them as `constexpr` instead.

src/utilities/utilities.hpp
```diff
@@ -28,6 +28,21 @@
...
+
+#ifdef XGEMM_EXEC
+extern bool tStrategyFlag;
+#endif
```
CNugteren: No global variables if you can avoid them. This is tuner related, so this is not a good place anyway. I'm sure you can get rid of the global variable and the `ifdef` as well :-)

src/utilities/utilities.hpp
```diff
@@ -206,8 +227,16 @@ struct Arguments {
...
+
+  int tStrategy = DEFAULT_STRATEGY;
+  size_t psoSwarmSize = DEFAULT_PSO_SWARM;
+  double psoInfG = DEFAULT_PSO_G;
+  double psoInfL = DEFAULT_PSO_L;
+  double psoInfR = DEFAULT_PSO_R;
 };
 
+
+
```
CNugteren: Try to keep the changes only to the minimum required. So no 2 extra empty lines if they are not really needed!

src/utilities/utilities.hpp
```diff
@@ -47,6 +62,12 @@ constexpr auto kUnknownError = -999;
 
 // =================================================================================================
 
+constexpr auto tStrategy = "strategy";
+constexpr auto psoSwarmSize = "psoSwarmSize";
+constexpr auto  psoInfG = "psoInfG";
+constexpr auto  psoInfL = "psoInfL";
+constexpr auto  psoInfR = "psoInfR";
+
```
CNugteren: Can you move these a bit lower under the section with the tuning-specific arguments and also follow their format: `kArgCamelCaseName` and the string itself as `lower_case_with_underscores`? But actually these are quite specific for this particular case of tuning, so perhaps we should move them out of the arguments structure and add them separately just for the tuners. That is a bit nicer IMHO, but it could have been done for some other existing cases as well...

src/utilities/utilities.hpp
```diff
@@ -206,8 +227,16 @@ struct Arguments {
   bool print_help = false;
   bool silent = false;
   bool no_abbrv = false;
+
+  int tStrategy = DEFAULT_STRATEGY;
+  size_t psoSwarmSize = DEFAULT_PSO_SWARM;
+  double psoInfG = DEFAULT_PSO_G;
+  double psoInfL = DEFAULT_PSO_L;
+  double psoInfR = DEFAULT_PSO_R;
```
CNugteren: See above, would also be nice to have these in the appropriate section with the designated comment. Or somehow just within the tuner itself if it's possible to do that nicely.

src/utilities/utilities.hpp
```diff
@@ -28,6 +28,21 @@
 #include "utilities/clblast_exceptions.hpp"
 #include "utilities/msvc.hpp"
 
+#define FULL_SEARCH_STRATEGY 0
+#define RANDOM_SEARCH_STRATEGY 1
+#define PSO_STRATEGY 2
+#define DVDT_STRATEGY 3
```
CNugteren: These should probably become C++ enum classes instead of defines.

src/tuning/tuning.hpp
```diff
@@ -93,13 +98,46 @@ void Tuner(int argc, char* argv[]) {
 
   // Use full-search to explore all parameter combinations or random-search to search only a part of
   // the parameter values. The fraction is set as a command-line argument.
-  if (args.fraction == 1.0 || args.fraction == 0.0) {
+  #ifdef XGEMM_EXEC
```
CNugteren: Can you try to make this a bit nicer? First of all no #ifdef's please! Second, try to integrate both cases in a nice way. Now there's 3 places where `UseFullSearch` is set! You can also just introduce one new command-line argument which is "use_pso" and is set to true per default and then have the parameters coded into the code? Not sure how much we would want to change them, but you probably know better than I do.

src/tuning/kernels/xgemm.cpp
```diff
@@ -190,10 +196,49 @@ void StartVariation(int argc, char *argv[]) {
...
+else
+{
+  printf("CK_TUNER_NUM_OF_STRATEGIES is not defined\n");     
+}
+
+printf("num_of_strategies : %d\n",num_of_strategies );
+
+if( DEFAULT_NUM_OF_STRATEGY != num_of_strategies )
+{
+  //FACCIO PRIMA LA FULL SEARCH
```
CNugteren: Italian comments in the code! Bella storia!

src/tuning/kernels/xgemm.cpp
```diff
@@ -190,10 +196,49 @@ void StartVariation(int argc, char *argv[]) {
...
+      StartVariation<2>(argc, argv);
+      break;
+    case PSO_STRATEGY:
+      StartVariation<3>(argc, argv);
+      break;
+    case DVDT_STRATEGY:
+      StartVariation<2>(argc, argv);
+      break;
+  }
+
```
CNugteren: Can we not keep it a bit simpler the way it was? And just replace random search by PSO and perhaps add an option to switch it back or so? This all seems overly complicated to me.

src/tuning/kernels/xgemm.cpp
```diff
@@ -54,7 +58,14 @@ class TuneXgemm {
   static size_t DefaultBatchCount() { return 1; } // N/A for this kernel
   static double DefaultFraction() { return (V==1) ? 1.0 : 512.0; } // test all or sample randomly
   static size_t DefaultNumRuns() { return 2; } // run every kernel this many times for averaging
-
+  static size_t DefaultNumSearchStragegy() { return 2; } // Full search and Random/PSO
+  static size_t DefaultSwarmSizePSO() { return 8; } 
+  static double DefaultInfluenceGlobalPSO(){ return 0.1; }
+  static double DefaultInfluenceLocalPSO(){ return 0.3; }
+  static double DefaultInfluenceRandomPSO(){ return 0.6; }
+  static size_t DefaultHeuristic(){ return size_t{3};} // PSO
```
CNugteren: It's a pity this comment "// PSO" is needed (and similar ones elsewhere). It would be much nicer to have a `scoped enum` instead of encoding these heuristic-choices with integers directly. That would make the code clearer and will no longer require these comments :-)

src/tuning/kernels/xgemm.cpp
```diff
@@ -190,10 +226,39 @@ void StartVariation(int argc, char *argv[]) {
...
+
+  auto command_line_args = clblast::RetrieveCommandLineArguments(argc, argv);
+  auto help = std::string{""};
+  auto heuristic_selected = clblast::GetArgument(command_line_args, help, clblast::kArgMultiSearchStrategy, 0); 
+  auto multi_search_strategy = clblast::GetArgument(command_line_args, help, clblast::kArgMultiSearchStrategy, 0);
+
+  if(multi_search_strategy){
+    StartVariation<1>(argc, argv);
+    StartVariation<2>(argc, argv);
+  }
```
CNugteren: This multi-search-strategy option: is that really needed? It complicates the code quite a lot and doesn't add much for the end user I would say. The end-user will just want to run the tuner as it is best to do it (e.g. using PSO if that's the best strategy) and don't care otherwise. For those that do experiments on the search strategy and other advanced users: they can now use the command-line option to switch between search strategies, right? I don't really see the use-case for doing two strategies at once.

src/tuning/kernels/xgemv.cpp
```diff
@@ -52,7 +52,14 @@ class TuneXgemv {
...
   static double DefaultFraction() { return 1.0; } // N/A for this kernel
   static size_t DefaultNumRuns() { return 10; } // run every kernel this many times for averaging
-
+  static size_t DefaultNumSearchStragegy() { return 1; } // N/A for this kernel
+  static size_t DefaultSwarmSizePSO() { return 8; } // N/A for this kernel
+  static double DefaultInfluenceGlobalPSO(){ return 0.1; }// N/A for this kernel
+  static double DefaultInfluenceLocalPSO(){ return 0.3; }// N/A for this kernel
+  static double DefaultInfluenceRandomPSO(){ return 0.6; }// N/A for this kernel
+  static size_t DefaultHeuristic(){ return size_t{0};} // Full search
+  static double DefaultMaxTempAnn(){ return 1.0;}// N/A for this kernel
```
CNugteren: This block of code is duplicated quite a few times. It is not so nice, mostly because these parameters are not relevant at all to this routine. But I think that's a problem of the design of the tuning infrastructure.... I think we should make a parent class to all these tuners and have some default settings in there. But I can do that outside of this PR.

mcian:
Hi, 
my apologies for not getting back to you sooner. I fixed the first request, I used the cltune::SearchMethod enum. 
For the second one, regarding the multistrategy, actually it allows to run one strategy (selecting which one) or to compare different strategies. 
In our experiments, where we usually work with eterogeneous  architectures, and we found the proposed solution more usable for respect to, for example, the original code every time runs the FullSearch and the Random. This is the main reason why I left the support for it.
I revert the code to the original one in the main function so the end-user will use it. If someone is interested in the multistategy can simple call the function TestHeuristic from main.

src/tuning/kernels/xgemm.cpp
```diff
@@ -63,7 +63,7 @@ class TuneXgemm {
   static double DefaultInfluenceGlobalPSO(){ return 0.1; }
   static double DefaultInfluenceLocalPSO(){ return 0.3; }
   static double DefaultInfluenceRandomPSO(){ return 0.6; }
-  static size_t DefaultHeuristic(){ return size_t{3};} // PSO
+  static size_t DefaultHeuristic(){ return (size_t)cltune::SearchMethod::PSO;} // PSO
```
CNugteren: Good idea to use the `cltune::SearchMethod` type! Now that we have that, can't we make the return type of this to be of that type as well (get rid of the `size_t`)? And then that can propagate further and can be directly set, right?

P.S. The comment is no longer needed here :-)

CNugteren:
OK, thanks for changing that. I've made a comment at the place where you use the `cltune::SearchMethod` type, but it doesn't show up here it seems. You can look for it in the code.

About the multi-strategy thing, OK, I understand that running the full-search every time is not nice. But in that case, can't you just remove the whole multi-strategy logic in the code and just comment out the `StartVariation<2>(argc, argv);` line if you don't want that for a particular experiment?

src/tuning/kernels/xgemm.cpp
mcian: Actually my first thought was to do like that, but then I realised that in order to propagate further the SearchMethod enum, we need to include the cltune header in utilities and, as side effect, it will be included everywhere. Do you think it is better? 
Yes, I will drop the comment

mcian:
> But in that case, can't you just remove the whole multi-strategy logic in the code and just comment out the StartVariation<2>(argc, argv); line if you don't want that for a particular experiment?

Yes but this will require to recompile the library. Sometimes, for example when we test a new product, it is useful to be able to switch between different strategies, configure and compare them by only providing some command line parameters. I think we can continue this talk and find a solution on a skype call, what do you think?

src/tuning/kernels/xgemm.cpp
CNugteren: Yes indeed, you are right. Perhaps I'll work on a tuner-specific utilities file and then we can do it later :-)

Small thing: could you use new C++11 style casts? So use `static_cast<size_t>(var_name)` instead? Thanks!

CNugteren:
> Yes but this will require to recompile the library. Sometimes, for example when we test a new product, it is useful to be able to switch between different strategies, configure and compare them by only providing some command line parameters

Yes, I understand, but isn't re-compilation also needed in the case of your [last commit|https://github.com/CNugteren/CLBlast/pull/173/commits/4adee608843b138f2edf15fbefdb61610edecb51]? In your experiments you would then just compile the tuner once with `StartVariation<2>(argc, argv);` removed and then you can use command-line arguments to select the strategy using `StartVariation<1>` which does support multiple strategies (but a single one at a time)?

mcian:
Of course now if you want to enable the multistrategy you need to recompile because it is turned off by default.  However when it is enable, you can simply switch between multiple search strategies at runtime; this is main benefit  of multistrategy in my opinion.
By the way, if you believe this is something useless for end-users, I will remove it from the pull request (we can keep it on our separate branch).

CNugteren:
I still don't understand. To be clear, with "mutlistrategy" I mean the option to run two strategies after each other in one run of the binary. I understand that you would like to experiment with different strategies and parameters without having to recompile, but that's still possible even if you run a single strategy, right? As I understand it, without the "multistrategy" option, you can just change the command-line arguments to select a different strategy and run the binary multiple times without having to compile.

mcian:
You right, you can choose between different strategies and parameters even if we remove the "multistrategy" option. The difference is that, with "multistrategy" we can easily support both your original version `StartVariation<1>(argc, argv); 
StartVariation<2>(argc, argv);` and the new version in which we can compare differrent strategies or simply select one. 
Otherwise, without the "multistrategy", in order to support both your original version (fullSearch and Random) and a command-line selected strategy, the easiest  solution is to comment one call to `StartVariation` when you want to run only one strategy at the time (and of course recompile).  
This is what I meant when I said you don't need to recompile; sorry if I was not clear in my previous comment ( I hope this will be more clear :-) )

CNugteren:
Sorry for being so picky, but I just see a lot of changes when I look at the diff, whereas they are not really needed in my opinion to make the search strategy an option in the command-line. I'm working with a quite large code-base and I'm trying to keep it as manageable and small as possible :-)

So indeed if the multi-search option would be a few extra lines, then OK. But right now it seems way more complicated to get it working. In my opinion all that's needed for the feature in this PR to work is changing the original code:
```
  if (args.fraction == 1.0 || args.fraction == 0.0) {
    tuner.UseFullSearch();
  }
  else {
    tuner.UseRandomSearch(1.0/args.fraction);
  }
```
Into something like:
```
 method = GetDefaultMethod();
 if (method == 0) { tuner.UseFullSearch; }
 if (method == 1) ...
  etc.
```
And then of course you'll need to set defaults for the search method and defaults for the PSO parameters as you already did. I don't see why we need the complexity of e.g. the `GetCurrentHeuristic` and `TestHeuristic` functions (both duplicated even!)?

mcian:
Hi, 
my apologies for not getting back to you sooner.
I got your point to keep the code as manageable and small as possible, so I removed the multistrategy option and all the related code. I also followed your suggestion about using static_cast.

CNugteren:
OK, thanks for the effort! It looks good now, I'l merge it in.

CNugteren:
@mcian I had to make some fixes (also changed to default random search to be conservative), because it didn't behave as expected. FYI, here is the commit with the changes: 6194d43.

mcian:
Hi Cedric,

I am sorry about that. I checked your commit and I saw there was a bug on
fraction. Thank you alert to me,  as soon as I have some spare time I will
check again the code and I will test it more extensively.
​

PR #174, opened 2017-07-17
Propose two different exhaustive search space for xgemm: light is a s…

mcian:
Hi Cedric,
 I propose two different exahustive search space for xgemm: the light one should provide on average good results but faster than the default one.
Best
Marco

### DIFF
--- a/src/tuning/kernels/xgemm_full.cpp
+++ b/src/tuning/kernels/xgemm_full.cpp
@@ -1,0 +1,200 @@
+
+// =================================================================================================
+// This file is part of the CLBlast project. The project is licensed under Apache Version 2.0. This
+// project loosely follows the Google C++ styleguide and uses a tab-size of two spaces and a max-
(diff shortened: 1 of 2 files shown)


CNugteren:
Hi Marco, I know it's been a while since you made this PR (we handled the other one first). But do you still think something like this is needed? If so, I propose to integrate it into the existing GEMM tuner as an option: I don't like all this code-duplication.

Note that the tuner architecture has changed (to prevent code duplication of default or unused settings), so this would have to be re-written anyway I am afraid...

mcian:
Hi Cedric,

to be honest I completely fogot this PR was still open. I agree with you,
right now it has to be re-written and it would be bette to integrate it as
an option.

I will close the PR.

Regards,
Marco

2017-10-08 11:06 GMT+02:00 Cedric Nugteren <notifications@github.com>:

PR #175, opened 2017-07-17
Add new threshold for ARM

mcian:
Hi Cedric,
I propose a specific threshold for ARM devices instead of the default one. This threshold seems to  most effective compared to the default one 

Best
Marco


CNugteren:
OK, merged it in. Not sure why the build of AppVeyor failed, can't find the issue in AppVeyor itself anywhere.

PR #206, opened 2017-10-23
Use GNUInstallDirs to determine install paths

matze:
The GNUInstallDirs module [1] provides variables that match the install directories for GNU Software and allows users to override them. Without hardcoding paths packagers can choose library paths according to distribution policies (i.e. lib, lib64, lib<arch>, ...).

[1] https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -307,27 +307,28 @@ if(MSVC)
 endif()
 
 # Installs the library
-install(TARGETS clblast EXPORT CLBlast DESTINATION lib)
-install(FILES include/clblast_half.h DESTINATION include)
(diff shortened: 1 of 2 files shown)


CNugteren:
OK, just tested on my system and `CMAKE_INSTALL_INCLUDEDIR` evaluates to `include` and `CMAKE_INSTALL_LIBDIR` evaluates to `lib`, so nothing really changed for me, which is a good thing. Also checked the requirement of CMake 2.8.5, which is below the current requirement of 2.8.11 so that's also fine.

Thanks for the addition!

PR #247, opened 2018-02-18
PyCLBlast: a Python interface for CLBlast based on PyOpenCL

CNugteren:
PyCLBlast: a Python interface for CLBlast based on PyOpenCL


scripts/generator/generator/pyclblast.py
```diff
@@ -0,0 +1,113 @@
+
+# This file is part of the CLBlast project. The project is licensed under Apache Version 2.0. This file follows the
+# PEP8 Python style guide and uses a max-width of 120 characters per line.
+#
+# Author(s):
+#   Cedric Nugteren <www.cedricnugteren.nl>
+
+NL = "\n"
```
fstakem: What about using os.linesep? A little wordy but you can alias it to NL. I suppose this only runs on linux but might be useful in future.

scripts/generator/generator/pyclblast.py
```diff
@@ -0,0 +1,113 @@
...
+SEPARATOR = "####################################################################################################"
+
+
+def to_np_dtype(flavour):
+    if flavour.precision_name == "S":
```
fstakem: I might have put that in a dict say like {'D': 'float64'}. Would possibly make the code a bit less redundant. The dict would have all the mappings not just the one I mentioned above.

scripts/generator/generator/pyclblast.py
```diff
@@ -0,0 +1,113 @@
...
+    raise RuntimeError("Could not convert flavour '%s' to numpy" % flavour.precision_name)
+
+
+def scalar_cython_conversion(scalar, flavour):
+    scalar_type = flavour.alpha_cl if scalar == "alpha" else flavour.beta_cl
+    if scalar_type == "float":
+        return "<cl_float>" + scalar
+    if scalar_type == "double":
+        return "<cl_double>" + scalar
+    if scalar_type in ["cl_float2", "float2"]:
```
fstakem: This one might be harder to put in a dict like above. If you wanted you could compare on the ending with 'value'.endswidth and construct the dict each time you enter the function.

scripts/generator/generator/pyclblast.py
```diff
@@ -0,0 +1,113 @@
...
+        result += SEPARATOR + NL
+        result += "# " + routine.description + ": " + routine.short_names() + NL
+        result += SEPARATOR + NL
+        result += NL
+
+        result += "cdef extern from \"clblast_c.h\":" + NL
+        np_dtypes = []
+        for flavour in routine.flavours:
+            if flavour.precision_name in ["S", "D", "C", "Z"]:
+                result += "    CLBlastStatusCode CLBlast" + flavour.name + routine.plain_name() + "("
```
fstakem: Just a suggestion. Maybe have a var for the tab size instead of the blank space at the beginning of each line. 
T = "    "
result += T + 'def*****'

fstakem: Sorry markdown is editing my space out of T var.

scripts/generator/generator/pyclblast.py
```diff
@@ -0,0 +1,113 @@
...
+            if option == "triangle":
+                result += "    triangle = CLBlastTriangleLower if lower_triangle else CLBlastTriangleUpper" + NL
+            if option == "diagonal":
+                result += "    diagonal = CLBlastDiagonalUnit if unit_diagonal else CLBlastDiagonalNonUnit" + NL
+
+        result += "" + NL
+        result += "    cdef CLBlastStatusCode err" + NL
+        if_prefix = ""
+        for flavour in routine.flavours:
+            if flavour.precision_name in ["S", "D", "C", "Z"]:
```
fstakem: If you define the dict I was suggesting above as a global in the file then you wouldn't have to have these vars here it would be a compare to see if key in dict.

scripts/generator/generator/routine.py
```diff
@@ -815,6 +815,38 @@ def arguments_doc(self):
...
+            if buf in self.buffers_matrix():
+                result.append(buf + "_ld")
+        for buf in buffers:
+            if buf in self.buffers_vector():
+                result.append(buf + "_inc = 1")
+        for scalar in self.scalars:
+            default = "1.0" if scalar == "alpha" else "0.0"
+            result.append(scalar + " = " + default)
+        for option in self.options:
+            if option == "a_transpose":
```
fstakem: Once again I would probably use a dict for this one so the if just checks the key in dict.

CNugteren:
Hi @fstakem, thanks for your remarks. They are indeed valid and I have fixed them now. I am treating this generator code a bit as a second-class citizen when compared to the main C++ code, but I should take better care. There is much more in there that can be improved actually :-)

PR #248, opened 2018-02-20
Fix of multiple duplicates in documentation

kpot:
Hi!
Just noticed multiple duplicates in lines describing arguments. Most likely the result of a wrong mass find/replace operation in text editor.

### DIFF
--- a/doc/clblast.md
+++ b/doc/clblast.md
@@ -254,8 +254,6 @@ Arguments to DOT:
 * `const size_t n`: Integer size argument. This value must be positive.
 * `cl_mem dot_buffer`: OpenCL buffer to store the output dot vector.
 * `const size_t dot_offset`: The offset in elements from the start of the output dot vector.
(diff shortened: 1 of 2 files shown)


CNugteren:
Thanks for contributing! Indeed, that's a duplicate that should be removed.

However, the documentation is generated using a Python script. So you should actually modify the script first, and then commit the updated script and the changes. Here's how to run the script:
```
python /path/to/clblast_root/scripts/generator/generator.py /path/to/clblast_root
```
e.g. when you are in the CLBlast folder:
```
python ./scripts/generator/generator.py .
```
And the issue seems to be [this duplicated line](https://github.com/CNugteren/CLBlast/blob/master/scripts/generator/generator/routine.py#L808) if I'm correct. Can you test that and update the PR if you have time? Otherwise I can also try it later.

kpot:
You were right about the line causing the duplicates. I've removed it and then tested by generating documentation again. It made an updated .md file which looks exactly like the manually corrected version I sent earlier. So I've updated the pull request.

PR #251, opened 2018-02-28
Added GEMM temp buffer C API

sivagnanamn:
Similar to CPP & CUDA, added C specific API for getting GEMM temp buffer size


CNugteren:
Thanks for this addition. There is one issue, which you might not know about (sorry for that), but some parts of these two files are generated automatically, since they are quite repetitive. So although manually adding to these files would be OK in theory, the next time you would run the generator it would remove your changes again. Here's how to run the generator from the root of CLBlast:
```
python scripts/generator/generator.py .
```
You'll need to make sure the generator doesn't change anything when running it with your changes in place.

You'll need to do two things:
* For `CLBlastGemmTempBufferSize` and the extra error code you can just change the amount of header and footer lines [here](https://github.com/CNugteren/CLBlast/blob/master/scripts/generator/generator.py#L52).
* For `CLBlastSgemmWithTempBuffer` you'll have to generate that function somehow automatically, or (probably easier) move it towards the end of the file and then add that as part of the footer.

Let me know if you have questions about that.

It would be nice if you can also modify the documentation for `GemmTempBufferSize` [here](https://github.com/CNugteren/CLBlast/blob/master/doc/api.md) in the same style as the others. Currently it says no C API is available.

sivagnanamn:
Sure, let me check the `generator.py` & API docs and get back :+1:

sivagnanamn:
@CNugteren I'll close this PR & create a new clean PR in some time. This seems a little messy.

TaihuLight:
The newest **API doc** [https://github.com/CNugteren/CLBlast/blob/master/doc/api.md](url) and **Supported routines overview** [https://github.com/CNugteren/CLBlast/blob/master/doc/routines.md](url) does not contain the description of  the added GEMM temp buffer C API, including the routines CLBlastXgemmWithTempBuffer() and CLBlastSGemmTempBufferSize().
Then, could you plan to update the API doc and Supported routines overview doc, to add the description of  APIs CLBlastXgemmWithTempBuffer() and CLBlastSGemmTempBufferSize(). Especially, which values are stored in the added temporary buffer? The values of Matrix C?
https://github.com/CNugteren/CLBlast/blob/1433dc67f17a94d1089291f4386d4fe668eb4a62/doc/api.md
@sivagnanamn @CNugteren

CNugteren:
In the API docs there is [GemmTempBufferSize](https://github.com/CNugteren/CLBlast/blob/master/doc/api.md#gemmtempbuffersize-retrieves-the-size-of-the-temporary-buffer-for-gemm-auxiliary-function), is this not what you are looking for?

If you are unclear what the temporary buffers are used for, then it might be good to read [this doc](https://github.com/CNugteren/CLBlast/blob/master/doc/details_gemm.md) and perhaps the linked CLBlast paper itself. In short, the temporary buffers are used to make certain assumptions of the in-direct GEMM kernel through. For example, if the GEMM kernel assumes `n` to be a multiple of 8, and it isn't, then a temporary buffer is used to pre-process in the input matrix and post-process the output matrix.

PR #253, opened 2018-03-02
Added C API for getting GEMM temp buffer size

sivagnanamn:
- Added C API for getting temp buffer size
- Updated generator.py header & footer lines 
- Update API doc


CNugteren:
Thanks!

CNugteren:
By the way, I had to fix some things regarding the generator (https://github.com/CNugteren/CLBlast/commit/bff64917bd5be9005c68e439517fbde06623e616). For now, that means part of the documentation is not there anymore, since it is (again) auto-generated. But I don't think that's a big issue.

PR #279, opened 2018-05-09
Update ci links to use doman names and build names instead of IP/id

umar456:
Updates the README badges to point to the domain name instead of
IP addresses. Also updates the names of the builds to the name
of the build instead of the id of the build.


CNugteren:
Thanks, much better indeed!

Issue #291, opened 2018-06-23
clblast_sample_cache_c segfaults


ranocha:
I've build CLBlast as described in the readme - using `-DVERBOSE=ON` after getting the segfault in order to get more details. I get
```
hendrik@i7-8700K:~/Downloads/CLBlast/build$ ./clblast_sample_cache_c Starting caching sample with an empty cache
[DEBUG] Searching database for kernel 'Xdot'
[DEBUG] Device type 'GPU'; vendor 'NVIDIA'
[DEBUG] Device name 'GeForce GTX 1070 Ti'; architecture 'SM6.1'
[DEBUG] Found architectures of vendor 'NVIDIA' and type 'GPU'
[DEBUG] Found devices of architecture type 'SM6.1'
[DEBUG] Found parameters for device type 'default'
[DEBUG] ASUM_Xdot_1024_32
[DEBUG] Compiling routine 'ASUM-32 (single)'
[DEBUG] Completed compilation in 0.87 ms
[DEBUG] Running kernel 'Xasum'
[DEBUG] Completed kernel in 0.06 ms
[DEBUG] Running kernel 'XasumEpilogue'
[DEBUG] Completed kernel in 0.01 ms
Completed routine with status 0 in 2.438 ms
[DEBUG] ASUM_Xdot_1024_32
Segmentation fault (core dumped)
hendrik@i7-8700K:~/Downloads/CLBlast/build$ 
```
Valgrind gives
```
hendrik@i7-8700K:~/Downloads/CLBlast/build$ valgrind ./clblast_sample_cache_c 
==20409== Memcheck, a memory error detector
==20409== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==20409== Using Valgrind-3.13.0 and LibVEX; rerun with -h for copyright info
==20409== Command: ./clblast_sample_cache_c
==20409== 
==20409== Warning: noted but unhandled ioctl 0x30000001 with no size/direction hints.
==20409==    This could cause spurious value errors to appear.
==20409==    See README_MISSING_SYSCALL_OR_IOCTL for guidance on writing a proper wrapper.
==20409== Warning: noted but unhandled ioctl 0x27 with no size/direction hints.
==20409==    This could cause spurious value errors to appear.
==20409==    See README_MISSING_SYSCALL_OR_IOCTL for guidance on writing a proper wrapper.
==20409== Warning: noted but unhandled ioctl 0x7ff with no size/direction hints.
[...]

PR #296, opened 2018-07-06
Eliminate a temporary Program object

alycm:
This was causing a crash for me because the temporary Program destructor called
clReleaseProgram on the cl_program with Program, and then clBuildProgram was
called on the same cl_program (belonging to the Program owned by the
shared_ptr, but it's the same cl_program).


alycm:
This relates to https://github.com/CNugteren/CLBlast/issues/291, see comment there.

CNugteren:
Ah, very good find! Yes, indeed, I already fixed the issue, but I think your approach is more desirable. I'll merge it in!

PR #297, opened 2018-07-11
inline PTX to support subgroup shuffle for Nvidia GPUs

tyler-utah:
Hi Cedric,

We spoke briefly at IWOCL about adding custom support for subgroups across some architectures, as currently only Intel is supported. I had a go at providing support for Nvidia GPUs using inline PTX:

https://docs.nvidia.com/cuda/inline-ptx-assembly/index.html

I've checked on my Intel HD5500 and Nvidia 940m and the code compiles and runs as expected.

I tried doing some sgemm tuning on Nvidia, but it is really difficult because the current constraints don't allow enable Nvidia subgroups. From xgemm_part1.opencl it requires (NWI != SUBGROUP_SIZE || MDIMC < SUBGROUP_SIZE). Because Nvidia has a subgroup size of 32, this isn't triggered as far as I could tell in any provided tuning. 

If you look at my first commit, I modified some of the tuning parameters to allow a small number of experiments around the Nvidia subgroups. I'm able to see a reliable performance difference of 92.5 GFLOPS vs 77.8 GFLOPS when subgroups are enabled on my 940m.

I would be interested in hearing your thoughts!

### DIFF
--- a/src/clpp11.hpp
+++ b/src/clpp11.hpp
@@ -44,6 +44,7 @@
 #include <numeric>   // std::accumulate
 #include <cstring>   // std::strlen
 #include <cstdio>    // fprintf, stderr
+#include "assert.h"
 
 // OpenCL
 #define CL_USE_DEPRECATED_OPENCL_1_1_APIS // to disable deprecation warnings
@@ -353,6 +354,12 @@ class Device {
     #endif
     return std::string{"SM"} + std::to_string(GetInfo<cl_uint>(CL_DEVICE_COMPUTE_CAPABILITY_MAJOR_NV)) +
            std::string{"."} + std::to_string(GetInfo<cl_uint>(CL_DEVICE_COMPUTE_CAPABILITY_MINOR_NV));
+  }
+
+  // Returns if the Nvidia chip is a Volta or later archicture (sm_70 or higher)
+  bool IsPostNVIDIAVolta() const {
+    assert(HasExtension("cl_nv_device_attribute_query"));
+    return GetInfo<cl_uint>(CL_DEVICE_COMPUTE_CAPABILITY_MAJOR_NV) >= 7;
   }
 
   // Retrieves the above extra information (if present)
--- a/src/clpp11.hpp
+++ b/src/clpp11.hpp
@@ -44,7 +44,7 @@
 #include <numeric>   // std::accumulate
 #include <cstring>   // std::strlen
 #include <cstdio>    // fprintf, stderr
-#include "assert.h"
+#include <assert.h>
 
 // OpenCL
 #define CL_USE_DEPRECATED_OPENCL_1_1_APIS // to disable deprecation warnings
--- a/src/kernels/level3/xgemm_part1.opencl
+++ b/src/kernels/level3/xgemm_part1.opencl
@@ -114,13 +114,29 @@ R"(
   #define GLOBAL_MEM_FENCE 0    // Global synchronisation barrier for potential better performance
 #endif
 
-// Intel subgroups (https://www.khronos.org/registry/OpenCL/extensions/intel/cl_intel_subgroups.txt)
-#ifndef USE_SUBGROUP_SHUFFLING
-  #define USE_SUBGROUP_SHUFFLING 0     // Optionally enables subgroup shuffling for Intel GPUs
+#ifndef NVIDIA_WARPS_AS_SUBGROUPS
+  #define NVIDIA_WARPS_AS_SUBGROUPS 0
 #endif
-#if USE_SUBGROUP_SHUFFLING == 1
+#ifndef NVIDIA_POST_VOLTA
+  #define NVIDIA_POST_VOLTA 0
+#endif
+#ifndef INTEL_SUBGROUP_EXTENSION
+  #define INTEL_SUBGROUP_EXTENSION 0
+#endif
+//#ifndef USE_SUBGROUP_SHUFFLING
+  #define USE_SUBGROUP_SHUFFLING 0     // Optionally enables subgroup shuffling for Intel GPUs
+//#endif
+
+// Intel subgroups (https://www.khronos.org/registry/OpenCL/extensions/intel/cl_intel_subgroups.txt)
+#if USE_SUBGROUP_SHUFFLING == 1 && INTEL_SUBGROUP_EXTENSION
   #define SUBGROUP_SIZE 8              // Assumes subgroup size is always 8 on Intel GPUs
 #endif
+
+// NVIDIA warps as subgroups using inline PTX (https://docs.nvidia.com/cuda/inline-ptx-assembly/index.html)
+#if USE_SUBGROUP_SHUFFLING == 1 && NVIDIA_WARPS_AS_SUBGROUPS
+  #define SUBGROUP_SIZE 32              // Assumes subgroup size is always 32 on NVIDIA GPUs
+#endif
+
 #if NWI != SUBGROUP_SIZE || MDIMC < SUBGROUP_SIZE
   #undef USE_SUBGROUP_SHUFFLING
   #define USE_SUBGROUP_SHUFFLING 0     // Disables subgroups in case the assumptions don't hold
--- a/src/kernels/level3/xgemm_part1.opencl
+++ b/src/kernels/level3/xgemm_part1.opencl
@@ -133,8 +133,7 @@ R"(
 #endif
 
 // NVIDIA warps as subgroups using inline PTX (https://docs.nvidia.com/cuda/inline-ptx-assembly/index.html)
-#if USE_SUBGROUP_SHUFFLING == 1 && (SUBGROUP_SHUFFLING_NVIDIA_PRE_VOLTA || \ 
-                                    SUBGROUP_SHUFFLING_NVIDIA_POST_VOLTA)
+#if USE_SUBGROUP_SHUFFLING == 1 && (SUBGROUP_SHUFFLING_NVIDIA_PRE_VOLTA || SUBGROUP_SHUFFLING_NVIDIA_POST_VOLTA)
   #define SUBGROUP_SIZE 32              // Assumes subgroup size is always 32 on NVIDIA GPUs
 #endif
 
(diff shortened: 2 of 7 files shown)


CNugteren:
Thanks for your contribution! Looks good at first glance, I'll have a more detailed look either tomorrow evening or this weekend. I'll run some performance tests myself as well on NVIDIA hardware.

Also, @StreamHPC might be interested to see this, as they are working on something similar.

src/clpp11.hpp
```diff
@@ -44,6 +44,7 @@
 #include <numeric>   // std::accumulate
 #include <cstring>   // std::strlen
 #include <cstdio>    // fprintf, stderr
+#include "assert.h"
```
CNugteren: ---> `<assert.h>`

src/clpp11.hpp
```diff
@@ -355,6 +356,12 @@ class Device {
            std::string{"."} + std::to_string(GetInfo<cl_uint>(CL_DEVICE_COMPUTE_CAPABILITY_MINOR_NV));
   }
 
+  // Returns if the Nvidia chip is a Volta or later archicture (sm_70 or higher)
+  bool IsPostNVIDIAVolta() const {
```
CNugteren: CLBlast also has a CUDA back-end, which works because every function in this file is also implemented in `cupp11.h`, so you'll have to mimic this behaviour with the same API in the other file as well to make the CUDA version still work.

src/kernels/level3/xgemm_part1.opencl
```diff
@@ -114,13 +114,29 @@ R"(
...
+  #define NVIDIA_WARPS_AS_SUBGROUPS 0
+#endif
+#ifndef NVIDIA_POST_VOLTA
+  #define NVIDIA_POST_VOLTA 0
 #endif
-#if USE_SUBGROUP_SHUFFLING == 1
+#ifndef INTEL_SUBGROUP_EXTENSION
+  #define INTEL_SUBGROUP_EXTENSION 0
+#endif
+//#ifndef USE_SUBGROUP_SHUFFLING
```
CNugteren: Why is this commented out? Now `USE_SUBGROUP_SHUFFLING` is always 0, or do I see it wrong?

src/utilities/compile.cpp
```diff
@@ -58,11 +58,27 @@ std::shared_ptr<Program> CompileFromSource(
     header_string += "#define GLOBAL_MEM_FENCE 1\n";
   }
 
-  // For Intel GPUs with subgroup support, use subgroup shuffling.
-  if (device.IsGPU() && device.HasExtension(kKhronosIntelSubgroups)) {
+  // For GPUs with subgroup support, use subgroup shuffling.
+  // Currently these are Intel via an extension and Nvidia using inline PTX (restricted to 32 bit)
+  if (device.IsGPU() && (device.HasExtension(kKhronosIntelSubgroups) ||
+                        (device.IsNVIDIA() && static_cast<int>(precision) == 32))) {
```
CNugteren: Better to formulate `static_cast<int>(precision) == 32` as `precision == Precision::kSingle`.

src/utilities/compile.cpp
```diff
@@ -58,11 +58,27 @@ std::shared_ptr<Program> CompileFromSource(
...
+    if (device.IsNVIDIA()) {
+      header_string += "#define NVIDIA_WARPS_AS_SUBGROUPS 1\n";
+
+      // Nvidia additionally needs to check pre or post volta due to new
+      // shuffle commands
+      if (device.IsPostNVIDIAVolta()) {
+        header_string += "#define NVIDIA_POST_VOLTA 1\n";
+      }
+    }
+    else if (device.HasExtension(kKhronosIntelSubgroups)) {
```
CNugteren: This logic is a bit doubled now. What if we split it, i.e. keep the original code and just add something, e.g.:
```
  // For Intel GPUs with subgroup support, use subgroup shuffling.
  if (device.IsGPU() && device.HasExtension(kKhronosIntelSubgroups)) {
    header_string += "#define USE_SUBGROUP_SHUFFLING 1\n";
    header_string += "#define INTEL_SUBGROUP_EXTENSION 1\n";
  }

  // For NVIDIA GPUs, use subgroup shuffling.
  if (device.IsGPU() && device.IsNVIDIA) {
    header_string += "#define USE_SUBGROUP_SHUFFLING 1\n";
    header_string += "#define NVIDIA_WARPS_AS_SUBGROUPS 1\n";
     ... // Volta stuff
  }

```
But now the `USE_SUBGROUP_SHUFFLING` is duplicated... oh well :-) Doesn't matter too much I guess, your solution is also fine.

src/kernels/level3/xgemm_part1.opencl
```diff
@@ -114,13 +114,29 @@ R"(
   #define GLOBAL_MEM_FENCE 0    // Global synchronisation barrier for potential better performance
 #endif
 
-// Intel subgroups (https://www.khronos.org/registry/OpenCL/extensions/intel/cl_intel_subgroups.txt)
-#ifndef USE_SUBGROUP_SHUFFLING
-  #define USE_SUBGROUP_SHUFFLING 0     // Optionally enables subgroup shuffling for Intel GPUs
+#ifndef NVIDIA_WARPS_AS_SUBGROUPS
```
CNugteren: I'm afraid to get lost in the subgroup defines.... could we rename them with a common start, e.g.:
```
SUBGROUP_SHUFFLING_INTEL
SUBGROUP_SHUFFLING_NVIDIA_PRE_VOLTA
SUBGROUP_SHUFFLING_NVIDIA_POST_VOLTA
```
And perhaps treat the two NVIDIA ones as separate things, either one of them will be set. I think this makes the host code clearer and also the kernel code.

src/clpp11.hpp
tyler-utah: Ah, I didn't notice! I can do that.

src/kernels/level3/xgemm_part1.opencl
tyler-utah: oops, I was doing that to get some performance numbers. I will add it back in. Good catch

src/kernels/level3/xgemm_part1.opencl
tyler-utah: Good idea. I'll work on this over the weekend. Thanks!

CNugteren:
One comment still about the code perhaps: You do have a backslash in one of the pre-processor macro's to break the line. Perhaps better to not do that and just keep it on one (longer) line? Not sure if all OpenCL compilers properly support that. And for some reason even g++ complains about it:
```
CLBlast-test/CLBlast/src/routines/level3/../../kernels/level3/xgemm_part1.opencl:137:76: warning: backslash and newline separated by space
```

I've then also done some tests on a V100 GPU. I changed the tuning parameters to match yours (I hope). Here is what I get without shuffling:
```
* Found 3 configuration(s)
* Parameters explored: GEMMK MWG NWG KWG MDIMC NDIMC MDIMA NDIMB KWI VWM VWN STRM STRN SA SB KREG 

|   ID | total |                                                                           param |       compiles |         time | GFLOPS |            status |
x------x-------x---------------------------------------------------------------------------------x----------------x--------------x--------x-------------------x
|  ref |     - |                                                                               - |             OK |      1.68 ms |      - |      reference OK |
x------x-------x---------------------------------------------------------------------------------x----------------x--------------x--------x-------------------x
|    1 |     3 |    1   64   64    1   64    2   64    2    1    1    1    0    0    0    0    1 |   OK      7 ms |      0.96 ms | 2237.7 |     results match |
|    2 |     3 |    1   64   64    1   64    2   64    2    1    1    1    0    0    0    0    2 |   OK      6 ms |      1.03 ms | 2094.9 |     results match |
|    3 |     3 |    1   64   64    1   64    2   64    2    1    1    1    0    0    0    0    4 |   OK      6 ms |      1.34 ms | 1599.6 |     results match |
x------x-------x---------------------------------------------------------------------------------x----------------x--------------x--------x-------------------x
```
And here is with your additions:
```
|   ID | total |                                                                           param |       compiles |         time | GFLOPS |            status |
x------x-------x---------------------------------------------------------------------------------x----------------x--------------x--------x-------------------x
|  ref |     - |                                                                               - |             OK |      1.66 ms |      - |      reference OK |
x------x-------x---------------------------------------------------------------------------------x----------------x--------------x--------x-------------------x
|    1 |     3 |    1   64   64    1   64    2   64    2    1    1    1    0    0    0    0    1 |   OK      8 ms |      0.90 ms | 2391.8 |     results match |
|    2 |     3 |    1   64   64    1   64    2   64    2    1    1    1    0    0    0    0    2 |   OK      7 ms |      0.88 ms | 2446.8 |     results match |
|    3 |     3 |    1   64   64    1   64    2   64    2    1    1    1    0    0    0    0    4 |   OK     10 ms |      0.89 ms | 2420.3 |     results match |
x------x-------x---------------------------------------------------------------------------------x----------------x--------------x--------x-------------------x
```
So quite some gain for the last case in particular. Can you confirm that I am doing the experiments correct, i.e. do you also get only 3 test cases with those settings?

tyler-utah:
I did the backslash so that the line doesn't exceed 100 characters, but I'll put it back; you make a good point about compiler support. 

And yes, I only got 3 test cases with the settings I played with. But I was going to ask: I'm not sure the current tuning configurations will ever trigger this feature. Do you think it would be worthwhile to add some new parameters to the gemm tuning? I don't understand all the options very well so I was wondering if you had some thoughts on this.

Thanks!

tyler-utah:
Sounds good to me; thanks for all the feedback!

If I have some time, maybe I'll look at the high-performing settings for Nvidia GPUs and see if shuffling can be incorporated.

I've also heard that shuffling for AMD can be incorporated in a similar way. I'm going to have a think about that too.

CNugteren:
> I've also heard that shuffling for AMD can be incorporated in a similar way. I'm going to have a think about that too.

From what I understood, Vincent's team at @StreamHPC already have something ready. Perhaps an idea to ask them first?

PR #332, opened 2018-11-09
Add im2colflip and col2imflip functions

vbkaisetsu:
This branch adds `xIM2COLFLIP` and `xCOL2IMFLIP` functions.

The convolution operation is done by `xCONVGEMM`, but `xIM2COL` is also necessary for containing additional operations. However, `xIM2COL` needs a flipped kernel for the convolution.
```
e.g.

X =
[ a b c ]
[ d e f ]
[ g h i ]

W =
[ 1 2 ]
[ 3 4 ]

conv2d(X, W) =
                      [ a b c ]
[ 4 3 2 1 ] . im2col( [ d e f ] )
                      [ g h i ]

** Note **
Convolution:
[ 4 3 ] => [ a b c ]
[ 2 1 ] => [ d e f ]
           [ g h i ]

Cross-correlation:
[ 1 2 ] => [ a b c ]
[ 3 4 ] => [ d e f ]
           [ g h i ]
```

Using `xIM2COLFLIP`, we can write the convolution without flipping a kernel manually.
```
conv2d(X, W) =
                          [ a b c ]
[ 1 2 3 4 ] . im2colflip( [ d e f ] )
                          [ g h i ]
```

These new functions are almost same as previous functions, but the output is swapped along kernel-w and kernel-h axes.


CNugteren:
Thanks for your contributions, but let's try to keep the API as limited as we can.

> These new functions are almost same as previous functions, but the output is swapped along kernel-w and kernel-h axes.

Since much of the code is the same as before, I would definitely consider re-using it. See for example what was done with routines like HEMV, re-using GEMV and the GEMV kernel-code as well through defines in the kernels, e.g. [here](https://github.com/CNugteren/CLBlast/blob/43e3f27254c4f7e4a0b332f5b88965c53c20bdd1/src/kernels/level2/xgemv.opencl#L54).

But maybe we can even consider expanding the current col2im and im2col API with an extra boolean argument 'swapped', 'flipped', or 'transposed' or whatever is the best name.

What do you think?

vbkaisetsu:
OK. I consider reusing the kernel code.

I think `flipped = true/false` or `target = kCrossCorrelation/kConvolution` is better if it takes an additional argument.
But I don't want to add a conditional branch in for-loops because it becomes slow.
How to implement it?

vbkaisetsu:
I tried adding `target` option to constructors of `Xim2col` and `Xcol2im` classes.
How do you think about it?

CNugteren:
But you'll need to pass an extra argument to the API of im2col and col2im, right? But, as you noticed, it is not easy to add a pre-processor define based on an API argument. So what you then can do is make 2 im2col kernels, one for each argument, and then choose the right kernel in the .cpp file (e.g. like [here](https://github.com/CNugteren/CLBlast/blob/master/src/routines/level1/xaxpy.cpp#L54)). Then you can call a common kernel function that accepts a regular argument, which you either hard-code to true or false, such that you don't need to duplicate all the kernel code. An example in a slightly different setting is [here](https://github.com/CNugteren/CLBlast/blob/master/src/kernels/level3/xgemm_direct_part3.opencl#L230): multiple kernel entry points, each calling the same kernel function with different argument. The compiler will in-line everything, so not to worry about using if-statements.

CNugteren:
Wow, you went quite deep into the structure of CLBlast, hopefully it didn't cost too much effort to understand things. Feel free to also ask questions to me in the future.

Anyway, I don't have much to comment on the PR, it all looks good and in the same style of the existing code. One thing that is now missing is the option to set this new `kernel_mode` argument from the CLBlast clients. Not super important, what is more important are the tests and you've done that :-)

PR #335, opened 2018-11-19
Remove unnecessary qualifier of inline function

vbkaisetsu:
This branch removes unnecessary qualifiers from `Xcol2im` inline function.
Some environments cannot compile it, and dump the following message: `'const bool' cannot be used as the type of a kernel parameter`


CNugteren:
I just saw that in the automated tests. Indeed, thanks for fixing :-)

Issue #267, opened 2018-03-20
Merge Im2Col with GEMM into a new convolution API


Ulfgard:
Hi,
This is a feature request.

The standard approach of first using im2col and then using gemm for a convolution has the disadvantage that one very easily runs out of memory. There is actually no reason to call im2col directly, ever(the three convolution primitives, convolution, derivative wrt the input and derivative wrt filter can be formulated as convolutions themselves). The approach that for example cuDNN follows is that the im2col matrix is only created implicitely inside the convolution-gemm kernel. instead of loading the part of the im2col matrix into local memory, it is created on the fly. It should be quite easy to do this in the framework provided by CLBlast.

Further, batch-processing of images should be made possible.

Ulfgard:
thanks! Please keep me in the loop. I would like to test it as soon as you have it working!

CNugteren:
By the way, do you have any link to a reference API for this, e.g. from cuDNN or some other library? Just want to make sure I implement this using the preferred interface.

Ulfgard:
Hi,

the cuDNN interface is horrible, see here for a somewhat minimal example:

https://gist.github.com/goldsborough/865e6717e64fbae75cdaf6c9914a130d
starting at around line 62. 

I am not a huge fan of many of the convolution extensions like striding or dilations, those things are what makes interfaces complex. I like different image ordering and i also like the distinction between convolution & cross-correlation, but i would rather put the latter in two different functions(the difference is flipping the filter and different padding for uneven pad-sizes). In terms of extensibility it is probably better to have a descriptor object, this way one could start minimal and expand from there.

kodonnell:
This is something I was planning to try too (though I'd need to learn a lot first), so I'm definitely in favor. The response by Scott Gray [here](https://petewarden.com/2015/04/20/why-gemm-is-at-the-heart-of-deep-learning/) might be useful - though it was a bit above my level, so I'm not sure.

@Ulfgard - my use case (like most others, I imagine) is convolutional neural nets etc. - I'd say kernel size, padding (generally just 'SAME' or 'VALID'), and stride are the most common arguments, and I'd favour using them. I haven't seen e.g. dilation used yet.

@CNugteren - if implementing a convolution kernel, would it pay to look at other approaches, as well as im2col + gemm? There are heaps lying round, which you're probably more aware of them than me, but libDNN and clDNN (which has multiple kernels, including winograd) come to mind.

Ulfgard:
[...]

PR #343, opened 2018-12-18
Fix single kernel version of convgemm

vbkaisetsu:
Related to #267 

This branch fixes the single kernel version of convgemm function, and enables `ConvGemmMethod::kSingleKernel` option.

It also adds a tuner that works on all parameters without errors.

### DIFF
--- a/src/kernels/levelx/xconvgemm_part1.opencl
+++ b/src/kernels/levelx/xconvgemm_part1.opencl
@@ -11,7 +11,6 @@
 // uses parameters from the direct GEMM kernel. This is the part with the loads from memory (1/2).
 // This uses "CONVGEMM_WITH_IM2COL" as a switch to select between direct convgemm or first running
 // the im2col kernel to create a 'col' temporary matrix.
-// TODO: Currently only works with 'CONVGEMM_WITH_IM2COL' set
 //
 // =================================================================================================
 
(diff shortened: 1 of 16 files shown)


CNugteren:
Great, thanks for working on this! I'll have a look at the PR tomorrow.

For next time though, please do let me know in the corresponding issue before you start working to prevent possible duplicate work (there's a change I could've picked it up as well!).

src/tuning/kernels/xconvgemm.hpp
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// Settings for this kernel (default command-line arguments)
+TunerDefaults XConvGemmGetTunerDefaults(const int) {
+  auto settings = TunerDefaults();
+  settings.options = {kArgChannels, kArgHeight, kArgWidth, kArgKernelH, kArgKernelW,
+                      kArgNumKernels, kArgBatchCount, kArgFraction};
+  settings.channels = 32;
+  settings.height = 66;
+  settings.width = 66;  // num_patches = 64x64 = 4096
```
CNugteren: Is 66 a sensible default here?

src/kernels/levelx/xconvgemm_part1.opencl
```diff
@@ -82,10 +87,17 @@ INLINE_FUNC real GlobalToLocalCheckedImage(const __global realMD* restrict image
       int idm = mg + GetGroupID0()*WGD;
       int idk = kg + kwg;
 
+      const int w_id = idm % output_w;
+      const int h_id = idm / output_w;
```
CNugteren: So this (and similar changes elsewhere) were the bugs in the kernel? Because I had everything implemented, but things didn't work well when running the tests & the tuner, so I suspected a couple of things were wrong.

src/tuning/kernels/xconvgemm.hpp
vbkaisetsu: `global_size` must be multiply of `local_size`, but the tuner directly sets `global_size` to `num_patches` without ceiling it.
Maybe, we also can adjust `global_size` in the tuning framework.

src/kernels/levelx/xconvgemm_part1.opencl
vbkaisetsu: Yes, it was a bug. `w_id` and `h_id` are also calculated in `Xconvgemm` function, but these two variables are skipped using `MWID`.

src/tuning/kernels/xconvgemm.hpp
CNugteren: Ah so this 66 is computed as 64 + 2 * padding, with the padding being one?

vbkaisetsu: It is more complicated.
```
// According to routines/levelx/xconvgemm.cpp
size_h = settings.height + 2 * pad_h
padding_h = dilation_h * (kernel_h - 1) + 1
output_h = (size_h - padding_h) / stride_h + 1

// XConvGemmSetArguments in tuning/kernels/xconvgemm.hpp
padding_h = 0
stride_h = 1
dilation_h = 1
kernel_h = 3

output_h = 64 -> settings.height = 66
```

tingxingdong:
@vbkaisetsu , just to clarify. Here the single kernel  convgemm, does the single mean  "having both im2col & gemm inside one kerenel"?

CNugteren:
Yes, a single kernel doing everything, thus no need for temporary memory. See #345 as well.

PR #345, opened 2018-12-31
Convolution with single kernel

CNugteren:
(for #267)

@vbkaisetsu: I've tested your branch and all seems fine. I then started making some modifications such that the tuner now actually stores its parameters in the CLBlast database, and the routine now uses the parameters from the database.

If I run `./clblast_client_xconvgemm -platform 2 -channels 32 -height 66 -width 66 -num_kernels 32 -batch_num 16` (same parameters as the default in the tuner), I get 46 GFLOPS on my Intel Skylake GPU.

If you want you can have a look at the last 3 commits to see the changes.


vbkaisetsu:
Thanks!

I ran 10,000 iterations of both `kWithIm2Col` and `kSingleKernel` with tuned kernels on GeForce GTX 1080 Ti.
Unfortunately, `kWithIm2Col` is faster than `kSingleKernel` (T_T)

```
=====================
channels:    3
height:    256
width:     256
kernel_h:    3
kernel_w:    3
pad_h:       0
pad_w:       0
stride_h:    1
stride_w:    1
dilation_h:  1
dilation_w:  1
num_kernels: 3
=====================
kWithIm2Col:   3099 [ms]
kSingleKernel: 4910 [ms]
```

CNugteren:
Thanks for testing. I originally suspected the two kernel approach would be faster in some cases, at least compared to a not super-optimised single kernel approach. But of course the main benefit is the great reduction in memory usage compared to the im2col approach. I can do some tests perhaps on other devices as well.

TaihuLight:
@vbkaisetsu @CNugteren 
How to test xconvgemm using kWithIm2Col and kSingleKernel with tuned kernels, I cant test it on Intel GPU 630 in I7-8700K and FT-2000+ CPU (2.3Ghz@64cores).

CNugteren:
If you want to test it, you can check out this branch, run the xconvgemm tuner, add the tuning results to the database (see the docs), re-compile, and then run the xconvgemm client, e.g. as I wrote about. Feel free to share the resulting JSON files of the tuner so I can include them in CLBlast.

TaihuLight:
I tuned the CLBlast on Intel GPU HD630 in I7-8700K and FT-2000plus CPU (2.3Ghz@64cores) . The script for executing tuners with default parameters and their generated *.JSON files as following:
file:///home/chengzi/Downloads/CLBlast-convolution/tuned-FT-2000Plus-CPU.tar.gz
file:///home/chengzi/Downloads/CLBlast-convolution/tuned-Intel-GPU-HD630.tar.gz
BTW,  could you share better script for tuning CLBlast?

CNugteren:
Now you are just running the regular tuners, which is something you can share in #1 if you want. This storyline here is about this PR with a new tuner and a modified convolution kernel. If you want to test, you could run that tuner as I mentioned above and share the results here.

And what do you mean with 'better'?

TaihuLight:
I have uploaded the tuned results on FT-2000+ to #1 .

PR #368, opened 2019-08-26
Fix out-of-bounds read/write in XhadFaster

etomzak:
Fix an error in XhadFaster where data would be written beyond the end of zgm.
The kernel loop assumed that there was always enough work for each thread to
process WPT items, but this was not enforced. It's possible to detect the
overflow with the "canary" buffer regions, but for SHAD, kCanarySize must be
~500 (much larger than the normal 127).

This commit may improve the performance of XhadFaster, since the kernel was
performing 2x work in some cases (once over real data, once over garbage).

Courtesy of Codeplay Software Ltd.

### DIFF
--- a/src/kernels/level1/xhad.opencl
+++ b/src/kernels/level1/xhad.opencl
@@ -97,10 +97,12 @@ void XhadFaster(const int n, const real_arg arg_alpha, const real_arg arg_beta,
   const real alpha = GetRealArg(arg_alpha);
   const real beta = GetRealArg(arg_beta);
 
-  if (get_global_id(0) < n / (VW)) {
+  const int num_desired_threads = n / (VW * WPT);
+
+  if (get_global_id(0) < num_desired_threads) {
     #pragma unroll
     for (int _w = 0; _w < WPT; _w += 1) {
-      const int id = _w*get_global_size(0) + get_global_id(0);
+      const int id = _w * num_desired_threads + get_global_id(0);
       realV xvalue = xgm[id];
       realV yvalue = ygm[id];
       realV zvalue = zgm[id];


etomzak:
I noticed that `XhadFastest` isn't being tested, because it would require a `kMatrixDims` of at least 1024. Adding that size globally slows the correctness tests down considerably, and it's beyond the scope of this PR, so I haven't made that change.

etomzak:
I can also confirm that oclgrind warns about the out-of-bounds accesses and that this PR makes the warnings go away.

src/kernels/level1/xhad.opencl
```diff
@@ -97,10 +97,10 @@ void XhadFaster(const int n, const real_arg arg_alpha, const real_arg arg_beta,
   const real alpha = GetRealArg(arg_alpha);
   const real beta = GetRealArg(arg_beta);
 
-  if (get_global_id(0) < n / (VW)) {
+  if (get_global_id(0) < (n / (VW * WPT))) {
```
CNugteren: Thanks for fixing this! I'm just trying to understand, it's been a while. If I understand it correctly the change here in the if-statement is basically the fix, right? The other change is just a different order of accessing the data, which might be faster on some devices but slower on others?

etomzak: Both of the changes are part of the fix.

`WPT` must be taken into account here, otherwise there are too many threads for the amount of data (`n`), and the threads do extra work past the ends of the buffers (because each thread wants to process `WPT` items).

The old `id` calculation creates an out-of-bounds access when `n < get_global_size(0)`. This is the case with, for example, the `SHAD` test when `xMatrixDims == 64`. It looks like it might be copy-pasta from `XhadFastest`.

I haven't done any further tuning on the `id` calculation, so there may well be a functionally equivalent access pattern that's faster. I've run a bunch of tests locally with different `xMatrixDims`, and I can see that the new `id` produces correct results and no out-of-bounds accesses.

TBH, I really struggled to understand the intent of the code, particularly the `WPT` parameter. It seems like a weird way to limit the number of workers, but I guess it's required for tuning(?). So TL;DR: I'm pretty sure my changes are functionally correct, but less sure they follow the intent of the kernel.

CNugteren: WPT is short for 'work-per-thread' and is a way to increase the amount of work a single thread does, which is a quite important tuning parameter indeed for many devices.

I took a bit more detailed look this time. The launch parameters make the global ID run till `n / (VW * WPT)`, rounded up to the work-group-size WGS. So the total extra work that might have been done is at most one WGS, which indeed for very small cases could be 2x as you stated earlier.

So the original code is this:
```
  if (get_global_id(0) < n / (VW)) {
    for (int _w = 0; _w < WPT; _w += 1) {
      const int id = _w*get_global_size(0) + get_global_id(0);
      ...
```
So here the WPT parameter is implemented in such a way that memory coalescing is enabled, e.g. with WPT=2 each thread does one item from the first half of the data, and one item from the second half. The bug I believe might be in the fact that the if-statement is missing the WPT parameter and that the global size might be larger than expected, right? So this would be OK I would say:
```
  int num_desired_threads = n / (VW * WPT);
  if (get_global_id(0) < num_desired_threads) {
    for (int _w = 0; _w < WPT; _w += 1) {
      const int id = _w*num_desired_threads + get_global_id(0);
      ...
```
Your solution basically implements that as well, but also changes the way the work is distributed, i.e.:
```
  int num_desired_threads = n / (VW * WPT);
  if (get_global_id(0) < num_desired_threads) {
    for (int _w = 0; _w < WPT; _w += 1) {
      const int id = get_global_id(0) * WPT + _w;
      ...
```
Could you try my suggestion with your test-code to see if that indeed works? Because I guess that would be much faster on GPUs in particular.

etomzak: I wasn't able to find any bugs with your suggested

```cpp
const int id = _w*num_desired_threads + get_global_id(0);
```

so I've updated the PR with the change.

CNugteren:
Thanks a lot for taking your time for this fix and doing all the testing :+1:

etomzak:
No problem. Happy to help.

PR #372, opened 2019-12-14
Reduced number of TestMatrix calls for the batched xgemm routines.

trantila:
Same level of guarantees to the previous looped test is given by testing the worst case scenario, only, reducing the TestMatrix call count from the number of batches to one, allowing for much higher batch counts without affecting performance.

### DIFF
--- a/src/routines/levelx/xgemmbatched.cpp
+++ b/src/routines/levelx/xgemmbatched.cpp
@@ -79,11 +79,9 @@ void XgemmBatched<T>::DoGemmBatched(const Layout layout, const Transpose a_trans
                              gemm_kernel_id);
 
   // Tests the matrices for validity
-  for (auto batch = size_t{0}; batch < batch_count; ++batch) {
-    TestMatrixA(a_one, a_two, a_buffer, a_offsets[batch], a_ld, false); // don't test for invalid LD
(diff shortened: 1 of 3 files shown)


trantila:
I witnessed some unexpected performance characteristics with this library a while back and it was figured out that the implementation of both xgemmbatched and xgemmstridedbatched tests the A, B and C matrices once for each batch. While this is generally speaking not a problem it does hamper the routines' scalability towards a case of high batch count (and small matrix dimensions) needlessly.

Took a stab at it and figured the `TestMatrix` calls made from the batched methods actually only "need" to test the worst case scenario with the matrix offset to reach the same level of guarantees, barring 64b integer overflows.

PS. The reason the `TestMatrix` calls performance so badly is due to the `CheckError` macro unconditionally calling `CLCudaAPIError::TrimCallString`, making a needless heap allocation if the checked status was good. This is seen through `TestMatrix` calling `Buffer::GetSize` which in turn applies the `CheckError` macro to `clGetMemObjectInfo`. This is nowhere near that big a thing on its own in my opinion though, but now you know.

CNugteren:
Thanks for finding this optimisation and explaining and implementing it correctly :-) I'll merge it in!

PR #392, opened 2020-06-05
Fix Program::GetIR to handle programs with multiple devices

### LINKED ISSUE #391: Overflow issue in Program::getIR with a queue created from context with multiple devices
The problem was initially reported [here](https://github.com/arrayfire/arrayfire/issues/2919).

The user was trying to use a custom context created out of two NVIDIA GPUs to use with ArrayFire. Though the issue occurred in the call to GEMM API, it should be a problem for other functions as well when a queue is created using a context that contains multiple devices.

The issue stems from [this line](https://github.com/CNugteren/CLBlast/blob/master/src/clpp11.hpp#L513) which assumes that the program object has only one associated device. The number of devices associated to the `cl_program` can be greater than one if the program is created using `clCreateProgramWithSource` which is passed a context that has more than one device.

I tried to come up with a minor patch (given below), but I don't think it is an efficient solution. Hence, raising this issue instead of a PR. The reason I think it may not be efficient is that the patch has an upstream effect of changing the signature of `Program::getIR` method and I am not full aware what consequences that will have. Moreover, the patch also broke tests as I expected since, I am only returning one of the binaries from `getIR()` in the patch.

```diff
diff --git a/src/clpp11.hpp b/src/clpp11.hpp
index aeb1498..2fcab4c 100644
--- a/src/clpp11.hpp
+++ b/src/clpp11.hpp
@@ -508,13 +508,27 @@ class Program {
 
   // Retrieves a binary or an intermediate representation of the compiled program
   std::string GetIR() const {
-    auto bytes = size_t{0};
-    CheckError(clGetProgramInfo(program_, CL_PROGRAM_BINARY_SIZES, sizeof(size_t), &bytes, nullptr));
-    auto result = std::string{};
-    result.resize(bytes);
-    auto result_ptr = result.data();
-    CheckError(clGetProgramInfo(program_, CL_PROGRAM_BINARIES, sizeof(char*), &result_ptr, nullptr));
-    return result;
+    cl_uint num_devices = 0;
+    CheckError(clGetProgramInfo(program_, CL_PROGRAM_NUM_DEVICES,
+                                sizeof(cl_uint), &num_devices, nullptr));
+
+    std::vector<size_t> binSizesInBytes(num_devices, 0);
+    CheckError(clGetProgramInfo(program_, CL_PROGRAM_BINARY_SIZES,
+                                num_devices * sizeof(size_t), binSizesInBytes.data(), nullptr));
+    std::vector<std::string> bins;
+    for (auto binSize: binSizesInBytes) {
+        auto result = std::string{};
+        result.resize(binSize);
+        bins.push_back(result);
+    }
+    std::vector<unsigned char*> out;
+    for (auto bin: bins) {
+        out.push_back((unsigned char*)bin.c_str());
+    }
+    //auto result_ptr = result.data();
+    CheckError(clGetProgramInfo(program_, CL_PROGRAM_BINARIES,
+                                num_devices * sizeof(char*), out.data(), nullptr));
+    return bins[0];
   }
 
   // Accessor to the private data-member
```

9prady9:
Fixes #391 

Local tests on a system with multiple NVIDIA GPUs passed. Regular run on a single GPU system also looks good.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@ Development version (next version)
 - Added batched routines to pyclblast
 - Added CLBLAST_VERSION_MAJOR/MINOR/PATCH defines in headers to store version numbering
 - Various minor fixes and enhancements
+- Fixed a bug in the caching when using a context with multiple devices
 
 Version 1.5.1
 - Implemented single-kernel version of convolution as GEMM
--- a/src/clpp11.hpp
+++ b/src/clpp11.hpp
@@ -519,6 +519,10 @@ class Program {
 
     auto bytes       = size_t{0};
     auto binSizeIter = size_t{0};
+    // Loop over the program binary sizes to find a binary whose size is > 0.
+    // The current logic assumes that there ever is only one valid program binary
+    // in a given cl_program. This should be the case unless the cl_program
+    // is built for all or a subset of devices associated to a given cl_program
     for (; binSizeIter < binSizesInBytes.size(); ++binSizeIter) {
         if (binSizesInBytes[binSizeIter] > 0) {
             bytes = binSizesInBytes[binSizeIter];
--- a/src/clpp11.hpp
+++ b/src/clpp11.hpp
@@ -509,12 +509,31 @@ class Program {
 
   // Retrieves a binary or an intermediate representation of the compiled program
   std::string GetIR() const {
-    auto bytes = size_t{0};
-    CheckError(clGetProgramInfo(program_, CL_PROGRAM_BINARY_SIZES, sizeof(size_t), &bytes, nullptr));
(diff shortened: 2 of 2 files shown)


CNugteren:
Thanks a lot, good to hear it works. Code looks good as well. Could you add two minor things to the PR and then I'll merge it:
1. Add a line in the file `CHANGELOG` with something like `- Fixed a bug in the caching when using a context with multiple devices`
2. Add a comment in the function you modified at the loop over bin-sizes indicating that we make the assumption here that there is only one valid program (which is valid for the current use of GetIR).

9prady9:
Will do, I shall send in the changes by tomorrow EOD.

PR #410, opened 2021-01-20
Use reference types to prevent unnecessary copying

jamesjer:
While building version 1.5.2 for Fedora, I noticed a number of compiler warnings of this form:
```
/builddir/build/BUILD/CLBlast-1.5.2/src/kernel_preprocessor.cpp:221:19: warning: loop variable 'array_name_map' creates a copy from type 'const std::pair<const std::__cxx11::basic_string<char>, long unsigned int>' [-Wrange-loop-construct]
  221 |   for (const auto array_name_map : arrays_to_registers) {  // only if marked to be promoted
      |                   ^~~~~~~~~~~~~~
/builddir/build/BUILD/CLBlast-1.5.2/src/kernel_preprocessor.cpp:221:19: note: use reference type to prevent copying
  221 |   for (const auto array_name_map : arrays_to_registers) {  // only if marked to be promoted
      |                   ^~~~~~~~~~~~~~
      |                   &
```

This patch uses references types as indicated to prevent unnecessary copying.

### DIFF
--- a/src/kernel_preprocessor.cpp
+++ b/src/kernel_preprocessor.cpp
@@ -218,7 +218,7 @@ void ArrayToRegister(std::string &source_line, const DefinesIntMap& defines,
                      const std::unordered_map<std::string, size_t>& arrays_to_registers,
                      const size_t num_brackets) {
 
-  for (const auto array_name_map : arrays_to_registers) {  // only if marked to be promoted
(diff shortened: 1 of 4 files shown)


CNugteren:
Thanks, makes sense.

Issue #415, opened 2021-02-27
CGEMM performance lower than SGEMM?


JishinMaster:
Hi,
Thanks again for this great library.

I used clblas a lot before switching to clBLAST, especially because it behaves better on other than AMD platforms due to the autotuning, and because some functions, including GEMM, have batch.
On an AMD GPU, the SGEMM performs better on clblas than on clBLAST, except on small matrices when using batch. On big matrices it achieves about 50 to 70% of theoretical bandwidth.
On CGEMM though, it does not behave as well. What I observed with clblas is that with CGEMM we can achieve better performance than with SGEMM (let's say 900GFlops/s on 4K by 4K matrices, instead of 700GFlops/s).
With clBLAST (I have tested the master, version 1.5.0, and version 1.4.1) if we can achieve 500Gflops/s with 4K SGEMM, we can only achive 200-250GFlops/s with CGEMM.
I have tested on both AMD devices with RoCM 4.0, and Nvidia hardware (old Kepler GT650m), and I observed the same behavior.

Is it to be expected?

CNugteren:
Good to hear you've discovered CLBlast!

First of all, CLBlast also uses tuning. However, it relies on tuning data from others to already get decent speed. So if there is a device similar to yours already in the database, it might be fast. If not, there might be something to gain. Second, there might be supporting smaller kernels around the main GEMM kernel that could be slow in your case. Third, the implementation might be slow for your specific matrix sizes and not for the default ones that it was tuned for.

We can test all of these 3 easily by running the tuner instead of running your own code. That way we'll only run a single GEMM kernel and will be able to find the best performance. So, first make sure you compile the library with `-DTUNERS=ON` passed to CMake. Then, run on both your devices the tuner and report the inspect the best reported GFLOPS values and report them here. Example run of the tuner: 
```
./clblast_tuner_xgemm --precision 3232 -m 512 -n 512 -k 512
```
(note that it runs sort of 4 times with 4 different kernel/search strategies)

JishinMaster:
Hi,
The performance I observed are obtained after following the clBLAST tuning guide (make alltuners + python script).

For good measure, I have relaunched the specific command you asked on both the GT650m (an ASUS laptop) and RX460 (a desktop).

- GT650m :
SGEMM :
1
* Got average result of 5.51 ms: 48.7 GFLOPS
* Found best result 1.61 ms: 167.0 GFLOPS
[...]

PR #416, opened 2021-03-07
set the correct flop count for xgemm

JishinMaster:
Set the correct flop count for xgemm (especially cgemm and zgemm).
Related to #415

### DIFF
--- a/src/tuning/kernels/xgemm.hpp
+++ b/src/tuning/kernels/xgemm.hpp
@@ -159,7 +159,13 @@ TunerSettings XgemmGetTunerSettings(const int V, const Arguments<T> &args) {
   }
 
   // Describes how to compute the performance metrics
-  settings.metric_amount = 2 * args.m * args.n * args.k;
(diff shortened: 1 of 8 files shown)


CNugteren:
Thanks for your pull request. I think it looks good. One note however: did you check for other routines with a similar computation? I guess the tuners/clients for other level3/x operations might have a similar formula that needs updating (e.g. triangular matrix mutliplication or batched-gemm operations).

JishinMaster:
I will look into it! :)

JishinMaster:
I upgraded the flops of some level3 and levelx functions.
I used the OpenBLAS for the flops formula (https://github.com/xianyi/OpenBLAS/tree/develop/benchmark).
It seems it would need to be extended further than level3 (axpy for instance).

CNugteren:
Thanks for your addition, looks good, let's merge it! The non-level-3 ops are typically measured in GB/s so that should not be an issue.

PR #417, opened 2021-04-30
Correct capitalization typo

gspr:
The CLBlastConfig.cmake file was installed to a directory named
CLBLast (notice second capital l), which can cause issues for CMake's
search path when looking for CLBlast on the system.

This commit also fixes other occurrences of the wrong capitalization,
all of it purely cosmetic (i.e. in comments).


CNugteren:
Thanks for catching this and fixing it!

Issue #422, opened 2021-08-11
DGEMM failures on Turing GPUs


umar456:
I am getting DGEMM failures on Turing GPUs. I have tested on the NVIDIA T4 and NVIDIA T2000. Here are the logs from my system:

```
* Running on OpenCL device 'Quadro T2000'.
* Starting tests for the 'DGEMM' routine. Legend:
   : -> Test produced correct results
   . -> Test returned the correct error code
   X -> Test produced incorrect results
   / -> Test returned an incorrect error code
   \ -> Test not executed: OpenCL-kernel compilation error
   o -> Test not executed: Unsupported precision
   - -> Test not completed: Reference CBLAS doesn't output error codes
* Testing with error margins of 0.5% (relative) and 0.001 (absolute)
* Testing 'regular behaviour' for '101 (row-major) 111 (regular) 111 (regular)':
   XXXXXXXX----XXXX---X---X-------XXXXXXXXX----XXXX---X---X-------X
   Error rate 24.43%: m=7 n=7 k=7 lda=7 ldb=7 ldc=7 offa=0 offb=0 offc=0 alpha=3.14 beta=3.14 
   Error rate 24.43%: m=7 n=7 k=7 lda=7 ldb=7 ldc=64 offa=0 offb=0 offc=0 alpha=3.14 beta=3.14 
   Error rate 23.86%: m=7 n=7 k=7 lda=7 ldb=64 ldc=7 offa=0 offb=0 offc=0 alpha=3.14 beta=3.14 
   Error rate 23.86%: m=7 n=7 k=7 lda=7 ldb=64 ldc=64 offa=0 offb=0 offc=0 alpha=3.14 beta=3.14 
   Error rate 24.43%: m=7 n=7 k=7 lda=64 ldb=7 ldc=7 offa=0 offb=0 offc=0 alpha=3.14 beta=3.14 
   Error rate 24.43%: m=7 n=7 k=7 lda=64 ldb=7 ldc=64 offa=0 offb=0 offc=0 alpha=3.14 beta=3.14 
   Error rate 24.43%: m=7 n=7 k=7 lda=64 ldb=64 ldc=7 offa=0 offb=0 offc=0 alpha=3.14 beta=3.14 
   Error rate 24.43%: m=7 n=7 k=7 lda=64 ldb=64 ldc=64 offa=0 offb=0 offc=0 alpha=3.14 beta=3.14 
[...]

PR #423, opened 2021-08-19
New tuning results for 1 Intel CPU and 5 NVIDIA GPUs

CNugteren:
Adds tuning results from @umar456 as supplied in #1 for:
- NVIDIA Tesla T4
- NVIDIA Tesla V100
- NVIDIA Quadro T2000
- NVIDIA Quadro GV100
 - NVIDIA A100
- Intel Core i9-9980HK

This should fix issue #422

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,8 @@
+Development version (next version)
+- Fix a correctness issue with DGEMM on SM 7.5 Turing GPUs
+- Various minor fixes and enhancements
+- Added tuned parameters for various devices (see doc/tuning.md)
+
 Version 1.5.2
 - Changed XAMAX/XAMIN to more likely return first rather than last min/max index, updated API docs
 - Added batched routines to pyclblast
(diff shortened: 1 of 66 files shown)


umar456:
I think it would be a good idea to mention limiting the clock speeds of the processors before performing the tuning tests. When I run the tuners on the CPUs and GPUs I run the following commands

Set the CPU frequency
```
sudo cpupower frequency-set -g performance
sudo cpupower frequency-set -u 3100
```

Set NVIDIA GPU frequency
```
sudo nvidia-smi -i <device id> -lgc <clock-speed>
```

You can get the possible frequencies for your NVIDIA GPU using the following command:
```
sudo nvidia-smi -i <device id> --query-supported-clocks=gr --format=csv
```

I suggest that you pick a clock speed that would be stable. Somewhere in the middle of the range of frequencies listed above.

CNugteren:
Sure, thanks. I've (almost) copied your comment into the docs.

PR #424, opened 2021-08-24
Update documentation to reflect CLBlast in Debian & Ubuntu

gspr:
Let the installation documentation reflect the fact that CLBlast is now in Debian and Ubuntu


CNugteren:
Thanks for the update. Can you perhaps also add a line that says something like: 'for older Ubuntu versions it can still be installed from the PPA at ...'?

PR #432, opened 2022-04-22
sum fix

justingra:


src/kernels/level1/xasum.opencl
```diff
@@ -93,7 +93,7 @@ void XasumEpilogue(const __global real* restrict input,
 
   // Computes the absolute value and stores the final result
   if (lid == 0) {
-    #if PRECISION == 3232 || PRECISION == 6464
+    #if (PRECISION == 3232 || PRECISION == 6464) && defined(ROUTINE_ASUM)
       asum[asum_offset].x = lm[0].x + lm[0].y; // the result is a non-complex number
     #else
       asum[asum_offset] = lm[0];
```
CNugteren, line 99: Could you elaborate on why this change is made? As far as I remember the old code, the final result of the ASUM and SUM routines is computed as a real value, summing both components of the complex number. This change that you make here now makes this true only in the case of the 'sum' routine, but in the case of the 'asum' routine it now returns a complex number instead.

BTW, did you run the tests afterwards? Because I guess if you change the definition here some tests might fail?

justingra: I ran the xasum tests and they passed.  I didn't see any for regular sums.  

I wanted the sum routine to provide me a complex sum when the requested routine was a regular sum.  It should be doing the opposite of what you said above.  It should provide a real value when absolute is requested and a complex value when it is not requested.  If this is not desired behavior then I will delete the PR.

CNugteren: > I ran the xasum tests and they passed. I didn't see any for regular sums.

Ah indeed, there are only xasum tests, no regular xsum tests. The reason for that is that actually the regular xsum is not part of the BLAS standard as far as I understand it.

>  It should be doing the opposite of what you said above

Yes you are right, it is the other way around of course.

> If this is not desired behavior then I will delete the PR.

What actually matters is the definition of course, and although I wrote this library, I'm not an expert on the definition of these routines. Let's look it up:
* According to my understanding of the [intel docs](https://www.intel.com/content/www/us/en/develop/documentation/onemkl-developer-reference-c/top/blas-and-sparse-blas-routines/blas-routines/blas-level-1-routines-and-functions/cblas-asum.html), asum returns the sum of the real and imaginary parts. And this is still correct, with or without your proposed change.
* As far as I can see, the non-absolute xsum routine is not part of a BLAS standard and not implemented by other libraries. So actually we can do what we want. And your suggestion makes sense: the user can always add those two values later if she/he wants, but the other way around is more tricky.

So in conclusion I'm happy to accept the PR. One downside is that it changes the definition of the routine. The question is how many users will be affected by this if they'll just upgrade to the next version and suddenly notice a change in behaviour. Could you at least add a one-line to the `CHANGELOG` file to indicate that this behaviour changed?

justingra: yep, will do

justingra: Alright, done, sorry for the delay!

PR #451, opened 2023-01-12
CLBlast modifications to address Qualcomm Adreno performance

quic-aangus:
## Summary of changes that address Qualcomm Adreno device performance:

As stated in the CLBlast README.md, CLBlast has "Performance issues on Qualcomm Adreno GPUs". These changes address the performance issues. After implementing these changes, we find that Adreno performance is improved by up to 2x across the CLBlast library.

### 1. Unique database entries for specific Adreno devices
- Adreno GPUs from generation to generation may have significantly different tuned kernel parameters. The original implementation of the CLBlast database does not distinguish between different Adreno versions. This modification populates the device architecture string to uniquely index Adreno GPU tuning parameters by architecture. 
- "AdrenoVersion()" method added to class "Device" in src/clpp11.hpp. Queries the Adreno version using the clGetDeviceInfo() function with the CL_DEVICE_OPENCL_C_VERSION string. 
image
- AdrenoVersion() is used in src/utilities/utilities.cpp by the function "GetDeviceArchitecture()". If the device is a Qualcomm device, then the info string returned by AdrenoVersion() is used to index the tuning results in the database. 
image

### 2. Uniquely indexed tuning results for Adreno a730 and a740
- Adreno tuning results using the new indexing described in 1
- "Adreno_730_tuning_results.zip" and "Adreno_740_tuning_results.zip" contain the json file tuning results and are included in the root directory of this pull request.

### 3. Toggle OpenCL kernel compilation options for Adreno
- Adreno supports compiler options that are implemented in CLBlast, but not toggled for Adreno devices. Compiler options are toggled for Adreno in src/utilities/compile.cpp

#### USE_INLINE_KEYWORD
image

#### USE_CL_MAD
image

#### GLOBAL_MEM_FENCE
image

### 4. New preprocessor directive RELAX_WORKGROUP_SIZE
- Specifying the **reqd_work_group_size()** attribute in OpenCL kernel definitions causes a significant performance drop for Qualcomm devices. This preprocessor directive removes the workgroup size requirement from the OpenCL kernels and its implementation is similar to other preprocessor directives defined in src/utilities/compile.cpp.
image
- In the common OpenCL code src/kernels/common.opencl we check for the preprocessor and set to off (require workgroup size) by default
image
- In the kernel definitions it looks like this:
image

### 5. Bug fix: handling of #undef in CLBlast loop unrolling and array-to-register mapping functions
- If USE_SUBGROUP_SHUFFLING was set, but the kernel parameters did not satisfy the conditions for subgroup shuffle, USE_SUBGROUP_SHUFFLING was not unset, resulting in kernel errors. Handling of #undef is added to the function PreprocessDefinesAndComments() in src/kernel_preprocessor.cpp.
image


CNugteren:
Thank you for your contribution. I'll try to find some time to review it this weekend.

src/kernels/level1/xamax.opencl
```diff
@@ -30,7 +30,11 @@ R"(
 // =================================================================================================
 
 // The main reduction kernel, performing the loading and the majority of the operation
-__kernel __attribute__((reqd_work_group_size(WGS1, 1, 1)))
+#if RELAX_WORKGROUP_SIZE == 1
+  __kernel
+#elif
```
CNugteren, line 35: Not sure what the difference is between an empty `#elif` and a normal `#else`, but the latter reads a bit nicer I think. Could you easily change all those `#elif`'s in `#else`'s?

src/utilities/compile.cpp
```diff
@@ -54,7 +54,7 @@ std::shared_ptr<Program> CompileFromSource(
 
   // For specific devices add a global synchronisation barrier to the GEMM kernel to optimize
   // performance through better cache behaviour
-  if (device.IsARM() && device.IsGPU()) {
+  if ((device.IsARM() && device.IsGPU()) || device.IsQualcomm()) {
```
CNugteren, line 57: Just to be sure (you probably know best), but is it sufficient to say `IsQualcomm` and not also make sure that it is a GPU? Not sure what other OpenCL Qualcomm devices there are, but they might not all require/want all these defines?

src/utilities/utilities.cpp
```diff
@@ -463,6 +463,9 @@ std::string GetDeviceArchitecture(const Device& device) {
     else if (device.HasExtension(kKhronosAttributesAMD)) {
       device_architecture = device.Name(); // Name is architecture for AMD APP and AMD ROCm
     }
+    else if (device.IsQualcomm()) { // queries the Adreno GPU architecture version
```
CNugteren, line 466: Similar to my other comment above, should we add a `device.isGPU()` to make sure this only applies to GPU Qualcomm devices? The `AdrenoVersion` function's name hints at the fact that this is GPU-only.

CNugteren (changes requested):
I left some small comments but overall the changes look good, and thanks for writing the detailed explanations of why the changes are necessary.

Just a few other general small comments:
* Could you add a sentence to the CHANGELOG file regarding your changes? And perhaps also remove that sentence about performance issues from the CLBlast README?
* Could you remove the two .zip files and attach them to the PR itself. Then I'll add them to the database
* Before merging everything in I'll make a few quick test runs on other hardware to make sure nothing breaks. I don't have much time these days so that might take a week or two.

Thanks!

src/kernels/level1/xamax.opencl
quic-aangus: Typo. `#elif` has been changed to `#else` in all the kernel files.

quic-aangus:
Thanks for the feedback!

* `#elif` --> `#else` in kernel files
* Included device type check in compile.cpp and utilities.cpp 
* CHANGELOG and README are updated. 
* Newer Adreno tuning results are removed from the code and included here.

[Adreno_730_tuning_results.zip](https://github.com/CNugteren/CLBlast/files/10437015/Adreno_730_tuning_results.zip)
[Adreno_740_tuning_results.zip](https://github.com/CNugteren/CLBlast/files/10437016/Adreno_740_tuning_results.zip)

CNugteren:
Thanks for making the changes, I think it looks good now. I will do some final tests before merging.

Any reason you closed the PR or was that a mistake?

quic-aangus:
@CNugteren We also have an android build script for CLBlast and some python functions for using CLBlast on android through adb (android debug bridge). They're used to push and execute the performance/tuning executables on device and fetch the tuning results. Is this something we should consider including? Or is it outside the scope of the library?

CNugteren:
I ran some basic tests on my system, everything seems to still work fine on a non-Qualcomm device :+1: I'll merge it in and afterwards add the tuning parameters to the database as well in a separate PR.

> @CNugteren We also have an android build script for CLBlast and some python functions for using CLBlast on android through adb (android debug bridge). They're used to push and execute the performance/tuning executables on device and fetch the tuning results. Is this something we should consider including? Or is it outside the scope of the library?

I guess this could be useful for other users as well. I assume these are just stand-alone scripts and functions or will they increase the complexity of existing code (e.g. CMakeLists)? If they are completely separate then I think it is fine to add them to the main CLBlast repository, since they won't be in anyone's way.

CNugteren (approved):

quic-aangus:
> I guess this could be useful for other users as well. I assume these are just stand-alone scripts and functions or will they increase the complexity of existing code (e.g. CMakeLists)? If they are completely separate then I think it is fine to add them to the main CLBlast repository, since they won't be in anyone's way.

The scripts for building, pushing, and tuning on device are standalone, but the python functions for using the performance executables on device require modification to the existing python code. For example, in scripts/benchmark/benchmark.py we made the following modification:

image

And in scripts/benchmark/utils.py we define run_binary_adb() to execute the binary through ADB and get_binary_result_adb() to fetch the results:

image

The modifications are simple, although not entirely separate from the existing code.

CNugteren:
> The modifications are simple, although not entirely separate from the existing code.

I think those changes are completely fine. In any case the benchmarking scripts are not part of the core CLBlast library, feel free to modify them as needed for Android.

Issue #454, opened 2023-03-22
Result of AMAX in different format to CBLAS


BadgerKing7:
The CLBlast AMAX routine computes the correct result of the operation, but returns it in the format of **<resulted_index> * incX** compared to other CBLAS implementations which return **<resulted_index>** only.

In the following examples, CLBlast has been tested against Intel's Math Kernel Library:
For float x = {1, 1.1, 1.2, 1.3, 1.4, 1.5, 1.6, 1.7, 1.8, 1.9, 2, 2.1, 2.2, 2.3, 2.4, 2.5, 2.6, 2.7, 2.8, 2.9, 3, 3.1, 3.2, 3.3, 3.4, 3.5, 3.6} with incX = 2,
CLBlast result = 26
MKL result = 13

For float-complex x = {(2.3,4.5), (2.4,4.6), (2.5,4.7), (2.6,4.8), (2.7,4.9), (2.8,5), (2.9,5.1), (3,5.2), (3.1,5.3), (3.2,5.4), (3.3,5.5), (3.4,5.6), (3.5,5.7)} with incX = 2,
CLBlast result = 12
MKL result = 6

CNugteren:
Thank you for reporting this issue.

Can you also reproduce this when building and running the `clblast_test_xamax` program? Because if I compare it against both AMD's clBlas and OpenBLAS (a CBLAS implementation) then I see no errors. So either the test coverage is not 100%, or the 'issue' is with the MKL library and not with other CBLAS libraries. Could you run this program yourself against both MKL and another CBLAS and see if the tests fail on your system?

BadgerKing7:
The tests included in the `clblast_test_xamax` executable are not failing on my system, but after some digging, I now believe the tests are not checking the correct things for this operation.

In my understanding, the final result of the AMAX routine should always be a single unsigned integer representing the index of the first element with the highest absolute value found in the input vector, regardless of the data type of the values in the input vector.
![image](https://user-images.githubusercontent.com/30077171/228831377-a63a8e76-8fc4-4a27-aadd-e30f1b7aa4b6.png) [link to lines](https://github.com/CNugteren/CLBlast/blob/master/test/correctness/testblas.cpp#L187)
[...]

PR #457, opened 2023-05-07
AMAX/AMIN integer testing and bug fixes

CNugteren:
This PR does three things. It makes the test infrastructure handle integer output, which uncovered two bugs. In summary:
- Fixed a bug that caused tests to not properly do integer-output testing (for XAMAX/XAMIN)
- Fixed a bug in XAMAX/XAMIN routines related to inadvertently including the increment and offset in the result
- Fixed a bug in XAMAX/XAMIN routines that would cause only the real part of a complex number to be taken into account

The issue was found in https://github.com/CNugteren/CLBlast/issues/454 by the user 'BadgerKing7'.

### DIFF
--- a/src/kernels/level1/xamax.opencl
+++ b/src/kernels/level1/xamax.opencl
@@ -55,7 +55,7 @@ void Xamax(const int n,
   while (id < n) {
     const int x_index = id*x_inc + x_offset;
     #if PRECISION == 3232 || PRECISION == 6464
-      singlereal x = xgm[x_index].x;
+      singlereal x = fabs(xgm[x_index].x) + fabs(xgm[x_index].y);
     #else
       singlereal x = xgm[x_index];
     #endif
--- a/src/kernels/level1/xamax.opencl
+++ b/src/kernels/level1/xamax.opencl
@@ -70,7 +70,7 @@ void Xamax(const int n,
     #endif
     if (x > max) {
       max = x;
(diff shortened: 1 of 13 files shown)


src/kernels/level1/xamax.opencl
```diff
@@ -70,7 +70,7 @@ void Xamax(const int n,
     #endif
     if (x > max) {
       max = x;
-      imax = id*x_inc + x_offset;
```
CNugteren, line 73: For reference: this is the bug fix for the indexing.

src/kernels/level1/xamax.opencl
```diff
@@ -70,7 +70,7 @@ void Xamax(const int n,
...
+      imax = id;
```
CNugteren, line 73: For reference: this is the bug fix for the indexing.

src/kernels/level1/xamax.opencl
```diff
@@ -55,7 +55,7 @@ void Xamax(const int n,
   while (id < n) {
     const int x_index = id*x_inc + x_offset;
     #if PRECISION == 3232 || PRECISION == 6464
-      singlereal x = xgm[x_index].x;
+      singlereal x = fabs(xgm[x_index].x) + fabs(xgm[x_index].y);
```
CNugteren, line 58: For reference: this is the bug fix for the complex values (float2/double2).

PR #477, opened 2023-05-22
Fix issues in Windows release script

CNugteren:
This fixes some issues seen in https://github.com/CNugteren/CLBlast/actions/runs/5039209358/jobs/9037203184 for Windows releases:
* The folder name with dots did not work, e.g. "clblast-1.6.0" created a folder named "clblast-1". Therefore the final .7z was empty.
* The build was done in debug mode instead of release mode, see https://stackoverflow.com/questions/19024259/how-to-change-the-build-type-to-release-mode-in-cmake

Fixes tested to work here: https://github.com/CNugteren/CLBlast/actions/runs/5049126014

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -79,16 +79,16 @@ jobs:
 
     - name: Run CMake
       run: |
-        mkdir ${{env.RELEASE_NAME}}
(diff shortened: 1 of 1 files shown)


LoganDark:
Please don't use 7z for these

image

there's no point in using 7z for such a small archive

CNugteren:
I think 7zip is fine to use for Windows releases, it is a free program that can be installed on a Windows machine, and will soon have [native support in Windows](https://www.bleepingcomputer.com/news/microsoft/windows-11-getting-native-support-for-7-zip-rar-and-gz-archives/).

But feel free to submit a patch if you strongly believe otherwise.

PR #490, opened 2023-06-15
Fix pointer error in `pyclblast` on ARM

yubrajbhoi:
Pointers can be negative in ARM, so using `size_t` causes an error like this:

```
$ python haxpy.py
# Setting up OpenCL
# Setting up Numpy arrays
# Setting up OpenCL arrays
# Example level-1 operation: AXPY
Traceback (most recent call last):
  File "/home/yubraj/others/yubrajgithub/CLBlast/src/pyclblast/samples/haxpy.py", line 35, in <module>
    pyclblast.axpy(queue, n, clx, cly, alpha=alpha_fp16)
  File "src/pyclblast.pyx", line 478, in pyclblast.axpy
OverflowError: can't convert negative value to size_t
```

Fixed by doing `:%s/<size_t>/<ptrdiff_t>/g`. After the patch:

```
$ python haxpy.py
# Setting up OpenCL
# Setting up Numpy arrays
# Setting up OpenCL arrays
# Example level-1 operation: AXPY
# Result for vector y: [2.5   4.332 6.168 8.   ]
# Expected result:     [2.5   4.332 6.168 8.   ]
```

Also, `platform` module was not imported in `setup.py`. Fixed that too.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 Development version (next version)
 - Added tuned parameters for various devices (see doc/tuning.md)
+- Fix pointer error in pyclblast on ARM
 
 Version 1.6.0
 - Modifications to improve performance on Qualcomm Adreno GPUs:
--- a/scripts/generator/generator/pyclblast.py
+++ b/scripts/generator/generator/pyclblast.py
@@ -121,10 +121,10 @@ def generate_pyx(routine):
 
         # Buffer transformation
         for buf in buffers:
-            result += indent + "cdef cl_mem " + buf + "_buffer = <cl_mem><size_t>" + buf + ".base_data.int_ptr" + NL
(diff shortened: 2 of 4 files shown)


CNugteren:
Thanks for your contribution. However, this .pyx file is generated using a Python script, so you'll have to make your change there as well, otherwise it will be undone on the next (manual) run of that script. You can do that here: https://github.com/CNugteren/CLBlast/blob/2b98c6a28c84fcb940e7712e726cacd3dab6fdb0/scripts/generator/generator/pyclblast.py#L50

yubrajbhoi:
@CNugteren Thanks for the feedback. I did not realise it was a generated file. I have applied the changes.

CNugteren (approved):

yubrajbhoi:
@CNugteren Is this a problem?

```
The following tests FAILED:
	 22 - clblast_test_xtrsv (SEGFAULT)
```

It seems `xtrsv` unit test is crashing on macOS. Not really related to `pyclblast`.

Issue #509, opened 2023-10-09
Pyclblast float16 scalar conversion


vathomass:
Hi,
First, thank for the nice software.
I am facing some issues using the floating point 16bit functions of `pyclblast`. In short, I am getting incorrect results. My driver is intel's OpenCL HD Graphics on windows and linux (through WSL2) platforms.
My take on the problem is that the type cast in the `cython` wrapper in incorrect, because it tries to cast floating point arguments to unsigned 16 bits integers, aka `cl_half` (see this issue in [khronos git](https://github.com/KhronosGroup/OpenCL-Headers/issues/105)).

To elaborate more, I have written a script (attached) that uses `axpy`, with the `alpha` parameter passed to the function as `float` or as `np.uint16`. In the latter case the results are numerically correct. I have done the same with the script of issue #334 (seems to me that the issue is related), which again returns the correct results in my PC with the `alpha` argument as  `np.uint16`. (I ran tests also with negative `alpha` values and the results are as expected).

My suggestion is to do the proper conversion in the `cython` wrapper, so the user does not have to do the conversion manually to use the floating point 16bit features. But I am unsure if all the OpenCL implementations do have this issue, namely treating `cl_half` as 16bit unsinged integer.

Best,
Thomas

[pyclblast16.zip](https://github.com/CNugteren/CLBlast/files/12843720/pyclblast16.zip)

CNugteren:
Sorry for my late reply.

Have you looked at the FP16 example [here](https://github.com/CNugteren/CLBlast/blob/master/src/pyclblast/samples/haxpy.py)? It uses the `float16` Numpy data-type, and also uses `pyclblast.float32_to_float16` for conversion of the `alpha` value.

vathomass:
Hi,

Sorry, I had not seen `pyclblast.float32_to_float16` because the link to this function in the discussion of #334 returned a 404. My bad, I should had searched the examples and/or the sources a little more. Indeed, running ```pyclblast.float32_to_float16(val)``` and `numpy.frombuffer(numpy.float16(val).tobytes(), np.uint16)[0]` return the exact same integer value (except for the case of overflow, where the first errors out and the second returns `NaN`).

[...]

PR #515, opened 2023-11-06
Floating point conversion in Cython wrapper.

vathomass:
This pull request is related to issue #509.

There are three changes:
1. Proper conversion of the floating point arguments to `cl_half`. This will preserve the fp value for the calculations. I see three ways that this can be done: a) use the native opencl header [`cl_half.h`](https://github.com/KhronosGroup/OpenCL-Headers/blob/main/CL/cl_half.h) which provides functions for conversion, b) use the clblast machinery for converting floats to halfs or c) use the numpy core library to do the conversion. My preference is option c) and this is the one I have implemented, as it does not add extra dependencies to the module.
2. Using generator.py script in windows add `CRLF` line endings, which `git` picks up as changes. For convenience, the line endings have been forced to be `LF`, independent of the platform.
3. Some routines expect an integer output (which is passed as `pyopencl` array class in the python implementation), like `amax` and `amin`. The cython wrapper expects these arguments as floats with the same accuracy as the other vector inputs. In the C side of the code these arguments are of type `size_t`. Python users have to convert the floating point values to integers to recover the value, by for example calling `np.frombuffer(val.get().tobytes(), np.uint32)[0]`. Moreover, with the fp16 these function error out with `RuntimeError: PyCLBlast: 'CLBlastXamax' failed: CLBlastInsufficientMemoryScalar: The unit-sized vector's OpenCL buffer is too small`. Changes in the generator script force these arguments to be unsigned integers with different size than the other routine inputs.

This example demonstrates the changes:
``` python
import numpy as np
import pyopencl as cl
from pyopencl.array import Array
import pyclblast

# Settings for this sample
dtype = np.float16
alpha = -55
n = 40

print("# Setting up OpenCL")
ctx = cl.create_some_context()
queue = cl.CommandQueue(ctx)

print("# Setting up Numpy arrays")
x = np.random.rand(n).astype(dtype=dtype)
y = np.random.rand(n).astype(dtype=dtype)

print("# Setting up OpenCL arrays")
clx = Array(queue, x.shape, x.dtype)
cly = Array(queue, y.shape, y.dtype)
clx.set(x)
cly.set(y)

print("# Example level-1 operation: AXPY")
pyclblast.axpy(queue, n, clx, cly, alpha=alpha)
print("# Result for vector y: %s" % cly.get())
print("# Expected result:     %s" % (alpha * x + y))

idx = Array(queue, (1, 1), np.uint32)
pyclblast.amax(queue, n=n, x=clx, imax=idx)
print("# Result for abs max:  %s" % idx.get()[0, 0])
print("# Expected result:     %s" % np.argmax(np.abs(x)))
```


CNugteren (approved):
Thank you very much for your contribution, and quite brave that you actually dug into my unreadable generator Python mess :-) Everything looks good, but just one comment: could you also update the `CHANGELOG` file and add 3 short entries that describe the three changes that you've made?

After this is merged I'll make a new release of the python bindings, so perhaps you can also already change `version="1.3.2"` into `version="1.4.0"? Thanks!

vathomass:
Hi,

Thanks for reviewing the code, happy to contribute. I have made the requested changes.

There is another issue with the `python` package, which I did not wanted to add to this PR, not to complicate things.
`Pyclblast` does not install on windows properly (it cannot find the `clblast.dll`, because windows ...). I have made some changes to trick it to work for me, but are not ready for PR (I have hard coded the paths for my PC). If I find time I will open a separate PR, but my time is limited for the next week. If you want to wait a week or two, it is OK, if not, it is OK.

Best
Thomas

CHANGELOG
```diff
@@ -2,6 +2,10 @@ Development version (next version)
 - Fix a bug in the pre-processor that would cause issues on Arm GPUs
 - Fix DLL install directory in mingw
 - Added tuned parameters for various devices (see doc/tuning.md)
+- Modifications to the python bindings (pyclblast)
+  * Convert float scalar values to cl_half for fp16 routines
+  * Amax/amin, max/min routines accept unsigned integer buffers for index
+  * Generator script always use LF endings, independent of the platform
```
CNugteren, line 8: This has nothing to do with pyclblast, right?
```suggestion
- Generator script now always use LF endings, independent of the platform
```

CNugteren (approved):

CNugteren:
Thanks for making the changes, I'll merge this in.

> f you want to wait a week or two, it is OK, if not, it is OK.

Take your time, I'll wait till the next PR before I make a new pyclblast release.

PR #519, opened 2023-12-12
Python module mutli-platform setup

vathomass:
Hi,

There were three issues when installing `pyclblast` in windows:
1. To compile the `python` module, you should provide to the linker the `clblast` and `opencl` include paths and the `clblast` library path. If not the module will not compile.
2. The `runtime_library_dirs` options in `distutils.extension.Extension`, used by the `setup.py` script, is not available in windows platforms (see [here](https://setuptools.pypa.io/en/latest/userguide/ext_modules.html)).
3. After a successful install, you get an `ImportError: DLL load failed` message when trying to import the module. This is probably related to this [issue](https://stackoverflow.com/questions/41365446/how-to-resolve-importerror-dll-load-failed-on-python).

This PR moves towards installing `pyclblast` in windows, without breaking the installation on others platforms. The `CLBlast` library uses `cmake` for building and installing, and also installs a `cmake` package for the library. Using `cmake` for building the `python` module also should address issues 1 and 2 (let `cmake`  take care of finding the `clblast` installation and setting up the compiler and linker flags). To solve issue 3, one approach (found in the internet) is to copy the dll along with the binary module to the installation folder.

Moreover, the PR switched from the `setup.py` style to the `pyproject.toml` style installation, hopefully for good.

I am a basic user of `cmake`, so I welcome any comments. I have tested the changes on windows and linux environments.

Best,
Thomas


src/pyclblast/README.md
```diff
@@ -53,6 +53,6 @@ How to release a new version on PyPi
 
 Following [the guide](https://packaging.python.org/tutorials/packaging-projects/), in essence doing (after changing the version number in `setup.py`):
 
-    python3 setup.py sdist bdist_wheel
+    python3 -m build
```
CNugteren, line 56: This did not work for me, while the old command does still work:
```
:~/CLBlast/src/pyclblast$ python3 -m build
python3: No module named build.__main__; 'build' is a package and cannot be directly executed
```
And if I remove that folder then I get `No module named build`. This is with the relatively old Python 3.7.

CNugteren (commented):
Thanks for your contribution. And apologies for the late review. I did leave one comment, but it isn't a super important one.

Otherwise it looks good to me. I've tested it on my Linux system and it builds fine and the tests run fine as well. Indeed, using the `pyproject.toml` should be the more modern version. I didn't test on Windows but I trust you on that, and given that it didn't work before this can't do much harm anyway ;-)

src/pyclblast/README.md
vathomass: Hi,

No worries for the time taken, it is a festive period after all.
I have updated the changelog file.

Regarding the error, it seems to me that you have not installed the [build](https://pypi.org/project/build/) python package. Running  `python3 -m pip install --upgrade build`  (or install it through your linux distro package manager) should make it work.
Keep in mind that the `python -m build` method for packaging, is the one currently described in the guide in the `README.md` [link](https://packaging.python.org/tutorials/packaging-projects/).
Let me know if the error persists after you install the `build` package.

One more unrelated thing, it would be good to add 'hints' (if you have any suggestion) for cmake to find the CLBlast library.

CNugteren (approved):

src/pyclblast/README.md
vathomass: I added some hints for CLBlast in the `CMakeLists.txt` file and updated the `README.md` to inform the user about them.

vathomass:
Hi,

I have also switched to [`scikit-build-core`](https://pypi.org/project/scikit-build-core/) for building the bindings, a module that uses CMake for building python packages. I assume it is better than the custom code I had added in the `setup.py` file.

Best
Thomas

CNugteren (approved):
Great, thanks. Looks really neat now!

Thanks again for your contribution, let's merge this.

PR #559, opened 2024-08-28
add check for test_matrix_a in gbmv,tbmv,hbmv and sbmv

andreimatraguna:


CNugteren (changes requested):
Unless I misunderstood the original issue, I don't think what you implemented here is correct. By changing `clblast.h` you are giving the option to the user to control whether or not the matrix A should be tested, but I thought the original issue was not about a new feature but about correctness: this matrix should be tested or it shouldn't, that is not up to the user.

Please correct me if I'm wrong.

Issue #574, opened 2025-06-24
Build fails with CMake 4.0


sergiodj:
Hi,

While building your software with CMake 4.0, I get:

```
2025/06/24 20:10:00 WARN CMake Error at CMakeLists.txt:12 (cmake_minimum_required):
2025/06/24 20:10:00 WARN   Compatibility with CMake < 3.5 has been removed from CMake.
2025/06/24 20:10:00 WARN
2025/06/24 20:10:00 INFO -- Configuring incomplete, errors occurred!
2025/06/24 20:10:00 WARN   Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
2025/06/24 20:10:00 WARN   to tell CMake that the project requires at least <min> but has been updated
2025/06/24 20:10:00 WARN   to work with policies introduced by <max> or earlier.
2025/06/24 20:10:00 WARN
2025/06/24 20:10:00 WARN   Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
```

This is mentioned in CMake 4.0's [release notes](https://cmake.org/cmake/help/v4.0/release/4.0.html#deprecated-and-removed-features).

PR #575, opened 2025-06-26
Fixed compatibility with CMake 4.0

CNugteren:
This should fix https://github.com/CNugteren/CLBlast/issues/574.


LunNova:
You can set a range like `2.8.11...3.10`. Newer CMakes pick the highest version within the range to set CMake policies.

Issue #584, opened 2025-07-29
Clang Format


EJain-Dev:
Can you please share a clang format file that fits the CLBlast format? Using the following lines for my Clang Format ends up formatting existing code:
```
BasedOnStyle: Google
ColumnLimit: 100
```

PR #590, opened 2025-08-11
Clang format: add to CI and apply to all files

CNugteren:
In https://github.com/CNugteren/CLBlast/issues/584 it was suggested to apply clang-format for a uniform code style. I myself am also in favor of clang-format and use it in all my more recent projects. However at the time of creating CLBlast no formatting tool was used, instead the [Google C++ style](https://google.github.io/styleguide/cppguide.html) was followed roughly.

This PR shows what would happen when applying clang-format to the existing code-base. A drawback is that a lot of git history is hidden/lost this way, because a lot of formatting changes are made. But perhaps that is acceptable.

Another issue is that the `scripts/generator` Python tool generates a few files, which are now not compliant to the code style and are re-formatted by clang-format. Also not a huge issue perhaps? The generator can also be modified to adhere a bit more to the clang-format style?

I'll try a few more style variants to see if I can reduce the diff while keeping a reasonably common code style.


CNugteren:
> reviewers could simply skip the few commits that involve clang format

It's not that easy, for example a simple git blame won't give you what you want. So there are definitely quite some drawbacks applying formatting at this later stage. Anyway, there are also advantages, so let's go through with it.

> more confusing to go through inconsistently formatted code

Hey, it wasn't that bad!

> Another issue is that the scripts/generator Python tool

I applied the easiest fixes. For the rest, I assume users of the generator tool run clang-format afterwards. This is now also automatically checked in CI.

I think we are good to go now.

Issue #597, opened 2025-08-13
Error while Running Copy Fast Tuner


EJain-Dev:
I get the following error from the current head of the master branch when running clblast_tuner_copy_fast.exe:

```* Options given/available:
    -platform 0 [=default]
    -device 0 [=default]
    -precision 32 (single) [=default]
    -m 1024 [=default]
    -n 1024 [=default]
    -alpha 2.00 [=default]
    -fraction 1.00 [=default]
    -runs 10 [=default]
    -max_l2_norm 0.00 [=default]

* Found 96 configuration(s)
* Parameters explored: COPY_DIMX COPY_DIMY COPY_WPT COPY_VW

|   ID | total |               param |      local      |      global     |       compiles |         time |   GB/s |            status |
x------x-------x---------------------x-----------------x-----------------x----------------x--------------x--------x-------------------x
|  ref |     - |                   - |       8       8 |    1024    1024 |OpenCL compiler error/warning:
1:300:48: error: use of undeclared identifier 'COPY_DIMX'
  __kernel __attribute__((reqd_work_group_size(COPY_DIMX, COPY_DIMY, 1)))
                                               ^
1:303:36: error: unknown type name 'realC'
                    __global const realC* restrict src,
                                   ^
1:304:30: error: unknown type name 'realC'
                    __global realC* dest,
                             ^
1:308:33: error: use of undeclared identifier 'COPY_WPT'
  for (int _w_one = 0; _w_one < COPY_WPT; _w_one += 1) {
                                ^
1:310:41: error: use of undeclared identifier 'COPY_WPT'
    const int id_two = (get_group_id(1)*COPY_WPT + _w_one) * COPY_DIMY + get_local_id(1);
                                        ^
1:310:62: error: use of undeclared identifier 'COPY_DIMY'
    const int id_two = (get_group_id(1)*COPY_WPT + _w_one) * COPY_DIMY + get_local_id(1);
                                                             ^
1:311:31: error: use of undeclared identifier 'COPY_VW'
    const int id = id_two*(ld/COPY_VW) + id_one;
                              ^
1:312:10: error: expected ';' after expression
    realC result;
         ^
         ;
1:312:5: error: use of undeclared identifier 'realC'
    realC result;
    ^
1:312:11: error: use of undeclared identifier 'result'
    realC result;
          ^
1:350:16: error: use of undeclared identifier 'result'
    dest[id] = result;;
               ^

* Exception caught with status -11 while running the reference, aborting
```

[...]

PR #598, opened 2025-08-13
Fix copy fast tuner

EJain-Dev:
This fixes issue #597.


EJain-Dev:
Those were the only kernels that failed to compile. Now I see some of the harms of integrating clang format into an existing codebase.

CNugteren:
Indeed, thanks for fixing, I fixed a few of these as well yesterday but forgot to look at the tuners.

Yeah I guess this is a bug in clang-format, it shouldn't re-order includes that are including directly in a string.

CNugteren (approved):

PR #599, opened 2025-08-14
Faster tuning

EJain-Dev:
This add multithreading to the tuners. When I was testing on my machine, compile times are significantly increased but because of thread overlapping, the overall runtime of the tuner was less.


EJain-Dev:
On my device, it is about a 2x speed up but I am working on trying to improve that before this PR gets merged.

EJain-Dev:
Now this provides a very good speed up if the number of threads is about the number of logical cores available. It also ensures that extra threads aren’t launched by setting a high value for the parameter.

EJain-Dev:
It seems that I accidentally launched the maximum number of threads instead of the minimum which was causing the hangs that you were experiencing.

As for the warnings you see, they should be ignored. The addPrintInfo function uses std::snprintf to calculate the size of the new concatenation after formatting. But to do this, it uses a char * of size 1 as an argument to snprintf first (this the compiler complains about) and then it allocates enough memory so that the next snprintf call succeeds. This does not cause undefined behavior since snprintf will not write if the size is not large enough.

src/tuning/tuning.cpp
```diff
@@ -89,6 +91,106 @@ void print_separator(const size_t parameters_size) {
...
+  std::condition_variable cv;
+  std::vector<size_t> global;
+  std::vector<size_t> local;
+};
+
+template <typename... Args>
+inline void addPrintInfo(std::string& str, const char* format, Args&&... args) {
+  char temp[] = "m";
+  auto size = std::snprintf(temp, 1, format, std::forward<Args>(args)...);
+  auto original_size = str.size();
```
CNugteren, line 108: Like this the warnings disappear (on my system at least):
```suggestion
  const auto size = std::snprintf(NULL, 0, format, std::forward<Args>(args)...);
  const auto original_size = str.size();
```

src/tuning/tuning.cpp
```diff
@@ -89,6 +91,106 @@ void print_separator(const size_t parameters_size) {
...
+    try {
+      auto configuration = configurations[config_id];
+      addPrintInfo(info.print_info, "| %4zu | %5zu |", config_id + 1, configurations.size());
+
+      for (const auto& parameter : settings.parameters) {
+        addPrintInfo(info.print_info, "%5zu", configuration.at(parameter.first));
+      }
+      addPrintInfo(info.print_info, " |");
+
+      // Sets the thread configuration
```
CNugteren, line 140: To avoid confusion with the C++ compilation thread:
```suggestion
      // Sets the OpenCL thread configuration
```

src/tuning/tuning.cpp
```diff
@@ -289,56 +392,36 @@ void Tuner(int argc, char* argv[], const int V, GetTunerDefaultsFunc GetTunerDef
...
-      const auto start_time = std::chrono::steady_clock::now();
-      auto compiler_options = std::vector<std::string>();
-      const auto program = CompileFromSource(kernel_source, args.precision, settings.kernel_name, device, context,
-                                             compiler_options, 0, true);
-      auto kernel = Kernel(program, settings.kernel_name);
-      const auto elapsed_time = std::chrono::steady_clock::now() - start_time;
-      const auto timing = std::chrono::duration<double, std::milli>(elapsed_time).count();
-      printf("   %sOK%s  %5.0lf ms |", kPrintSuccess.c_str(), kPrintEnd.c_str(), timing);
-
       // Runs the kernel
```
CNugteren, line 425: Just making sure: running the kernel doesn't run in the parallel loop, right? Because on some devices launching multiple OpenCL kernels at the same time might influence performance numbers.

src/tuning/tuning.cpp
```diff
@@ -119,6 +221,7 @@ void Tuner(int argc, char* argv[], const int V, GetTunerDefaultsFunc GetTunerDef
   args.device_id =
       GetArgument(command_line_args, help, kArgDevice, ConvertArgument(std::getenv("CLBLAST_DEVICE"), size_t{0}));
   args.precision = GetArgument(command_line_args, help, kArgPrecision, Precision::kSingle);
+  args.threads = GetArgument(command_line_args, help, kArgNumThreads, size_t{1});
```
CNugteren, line 224: Good idea to keep this thread count at 1 as default. Because on some systems, the host and the device might be the same: we don't want parallel compilation to influence the kernel run-time!

CNugteren (commented):
Thanks for fixing it, it runs now on my system. And indeed, it is very helpful. Here is the output of the Linux `time` command for an example run of compiling and launching 100 GEMM kernels:
```
./clblast_tuner_xgemm --threads 1  58.02s user 5.70s system 119% cpu 53.415 total
./clblast_tuner_xgemm --threads 2  67.23s user 7.63s system 240% cpu 31.123 total
```
:rocket: 

Apart from the comments I left above, can you also add a note to the changelog? And maybe also write some info about this in `doc/tuning.md`? For example that this option exists in the first place, that it is recommended to put it to the number of CPU cores that you have in your system, that using more might degrade speed, and that using a single thread is the safest, in particular if you are tuning for a CPU.

CHANGELOG
```diff
@@ -2,6 +2,7 @@ Development version (next version)
 - Fixed compatibility with CMake 4.0
 - Applied clang-format code style formatting
 - Added tuned parameters for many devices (see doc/tuning.md)
+- Enabled drastic improvements for tuning times via multithreading (see doc/tuning.md)
```
CNugteren, line 5: Let's make this a bit less of an advertisement and more concrete:
```suggestion
- Enabled parallel kernel compilation for faster kernel tuning (see doc/tuning.md)
```

src/tuning/tuning.cpp
```diff
@@ -119,6 +220,7 @@ void Tuner(int argc, char* argv[], const int V, GetTunerDefaultsFunc GetTunerDef
   args.device_id =
       GetArgument(command_line_args, help, kArgDevice, ConvertArgument(std::getenv("CLBLAST_DEVICE"), size_t{0}));
   args.precision = GetArgument(command_line_args, help, kArgPrecision, Precision::kSingle);
+  args.threads = GetArgument(command_line_args, help, kArgNumThreads, size_t{1}) - 1;
```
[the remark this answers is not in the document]
EJain-Dev: It was to make it extra threads, I did the renaming.

src/tuning/tuning.cpp
```diff
@@ -89,6 +91,105 @@ void print_separator(const size_t parameters_size) {
...
+template <typename... Args>
+inline void addPrintInfo(std::string& str, const char* format, Args&&... args) {
+  auto size = std::snprintf(nullptr, 0, format, std::forward<Args>(args)...);
+  auto original_size = str.size();
+  str.resize(original_size + size);
+  std::snprintf(&str[original_size], size + 1, format, std::forward<Args>(args)...);
+}
+
+template <typename T>
+void tuningThread(std::vector<ThreadInfo>& infos, const std::vector<clblast::Configuration>& configurations, size_t id,
```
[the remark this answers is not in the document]
EJain-Dev: Implemented.

Issue #592, opened 2025-08-13
Fixing Deprecation Warning


EJain-Dev:
clpp11.hpp uses clCreateCommandQueue which is throwing a lot of deprecation warnings, why not just use an `#ifdef` to only use clCreateCommandQueue if the OpenCL Target Version is less than 200 and otherwise replace with clCreateCommandQueueWithProperties?

EDIT: It seems that deprecation warnings should be disabled via the macro, but build still show it, why is that?

EDIT 2: I think I’ve traced the source of the deprecation warnings. `clblast.h` is include via `utilities.hpp` which is included by almost every file. `utilities.hpp` also includes `clpp11.hpp` but `clblast.h` is included first. Then `clpp11.hpp` is included but because `clblast.h` directly includes OpenCL headers and does not define target version, `clpp11.hpp` redefines it (1 of the warnings) and then it leads to the deprecation warning.

EDIT 3: Proposed Solution: Clang Format will always reorder the includes in this manner unless 
1. Commenting in between includes complicated
2.  `clblast.h` can directly include `clpp11.hpp`
3. Completely reworking the include graph (not ideal)

EJain-Dev:
Pull request #604 fixes these warnings.

PR #604, opened 2025-08-14
Fix Compilation Warnings

EJain-Dev:
Fixed the OpenCL related warnings mentioned in Issue #592.


EJain-Dev:
What is the lint failure here? Nothing changed other than the `#ifndef, #elif, #endif` block.

CNugteren:
> What is the lint failure here

You can run those steps from the yml file locally to check yourself. But what's happening in this linting test is that we first run the generator, then clang-format, and then check that there is no diff with your code. Apparently there is, which means you'll need to update the generator script as well.

Issue #602, opened 2025-08-14
Improving Kernel Performance on Certain Devices


EJainDev:
Some kernels make default assumptions (a parameter is 0, stride, offset, etc.), why not use `cl_khr_expect_assume` (if supported) to make them explicit for some compilers?

PR #610, opened 2025-08-23
cl_khr_expect_assume Extension Support

EJainDev:
This enables optimizations by the compiler based on the __has_builtin which is currently supported by the clang compiler for OpenCL. Other compilers may also support this feature. This was mentioned in issue #602.

### DIFF
--- a/doc/tuning.md
+++ b/doc/tuning.md
@@ -216,7 +216,23 @@ The kernels `gemm` and `gemm_direct` have too many parameters to explore. Theref
 
 There are also several routine-level tuners. They tune inter-kernel parameters and should only be run after the kernels are tuned. However, they do automatically pick up kernel tuning results from the current folder if there are any. An example is the GEMM routine tuner, which determines when to use the direct or the in-direct GEMM kernel.
 
-The tuners also proivide a `-threads` option allowing you to control how many threads are used for OpenCL kernel compilation (not for actually executing the kernels). It defaults to running the single threaded version with 1 thread but more can be specified via the parameter. It is recommended to use the same amount of threads as CPU cores to maximize performance. More threads may hurt or improve performance. It is also the safest option to use the default of 1 thread.
+Common Tuning Parameters
+-------------
+
+The tuners provide a few common parameters that are shared by each tuner:
+1. **Precision** -- This is the precision type to train for, valid options are:
+    * 16 for real 16 bit loating point numbers
+    * 32 for real 32 bit loating point numbers
+    * 64 for real 64 bit floating point numbers
+    * 3232 for complex 32 bit floating point numbers
+    * 6464 for complex 64 bit floating point numbers
+2. **Platform** -- The OpenCL platform to use
+3. **Device** -- The OpenCL device to use
+4. **Fraction** -- The fraction of a larger search space to explore when running the tuners. A value of 100 is equal to 1% and so 10000 is equal to the whole search space (100%)
+5. **Threads** -- The number of threads that the tuner should use for compiling the OpenCL kernels. This does NOT run the OpenCL kernels with multithreading, it only compiles them with multithreading. 1 is the minimum and the recommended amount is the number of cores present in the CPU (with hyperthreading) and more may hurt prformance. It is safer to use 1 thread, espescially for tuning CPUs. If you specify more threads than there are kernels, it will simply use the number of kernels as the number of threads.
+
+Running All Tuners For All Precisions
+-------------
 
 Here are all the tuners included in the `make alltuners` target (in the same order) with all their precision arguments:
 


src/kernels/level2/xgemv_fast.opencl
```diff
@@ -98,6 +98,14 @@ void XgemvFast(const int m, const int n,
                __global real* ygm, const int y_offset, const int y_inc,
                const int do_conjugate, const int parameter,
                const int kl_unused, const int ku_unused) {
+#if __has_builtin(__builtin_assume)
+  __builtin_assume(m % WGS2 == 0);
+  __builtin_assume(n % WGS2 == 0);
+  __builtin_assume(a_ld % VW2 == 0);
+  __builtin_assume(a == 0);
+  __builtin_assume(a_rotated == 0);
+  __builtin_assume(do_conjugate == 0);
```
CNugteren, line 107: These aren't really used in this kernel, so I don't think this is useful. In fact, it might hurt performance because now these values are 'used' suddenly?

CNugteren (commented):
Did you measure any speedups? If so, for which kernels? On which devices did you test? Can you share tests logs for platforms both with and without this extension to make sure things still work?

src/kernels/level2/xgemv_fast.opencl
EJain-Dev: I don't think it would have hurt performance, but just to be sure I have removed it.

EJain-Dev:
Here are the speed ups on my device on GEMV since it is the most complicated kernel I edited. It is a decent speed up of about 3-20%. (I checked performance through tuners, I ran once on master branch and once on this branch):

Run 1 (master): <--------- (Doesn't matter, no edits were done)
* Got average result of 1.22 ms: 13.8 GB/s
* Found best result 0.87 ms: 19.4 GB/s
* Best parameters: PRECISION=32 WGS1=128 WPT1=1

Run 1 (branch): <--------- (Doesn't matter, no edits were done)
* Got average result of 1.28 ms: 13.2 GB/s
* Found best result 0.88 ms: 19.1 GB/s
* Best parameters: PRECISION=32 WGS1=256 WPT1=1

Run 2 (master):
* Got average result of 1.06 ms: 15.8 GB/s
* Found best result 0.85 ms: 19.8 GB/s
* Best parameters: PRECISION=32 VW2=4 WGS2=16 WPT2=4

Run 2 (branch):
* Got average result of 1.03 ms: 16.3 GB/s
* Found best result 0.83 ms: 20.2 GB/s
* Best parameters: PRECISION=32 VW2=1 WGS2=64 WPT2=2

Run 3 (master):
* Got average result of 7.65 ms: 2.2 GB/s
* Found best result 1.40 ms: 12.0 GB/s
* Best parameters: PRECISION=32 VW3=2 WGS3=64 WPT3=32

Run 3 (branch):
* Got average result of 5.85 ms: 2.9 GB/s
* Found best result 1.06 ms: 15.8 GB/s
* Best parameters: PRECISION=32 VW3=2 WGS3=64 WPT3=32

CNugteren:
Thanks for sharing. However, I would not involve the tuner: just keep the tuning parameters as is. The reason is that the tuner isn't perfect due to measurement noise, so a re-run of the tuner might give you different parameters that work worse or better.

So I would simply run e.g. `clblast_client_gemv` (and other related clients) twice: once on `master` and once on this branch. And then do the same for a device that doesn't support this extension, so be safe. And then run the test suite as well of course.

EJain-Dev:
Here are the results for xgemv on this PR from the client after tuning. From what I see, there is a slight improvement:
```
* Options given/available:
    -m 512 [=default]
    -n 512 [=default]
    -layout 101 (row-major) [=default]
    -transA 111 (regular) [=default]
    -lda 512 [=default]
    -incx 1 [=default]
    -incy 1 [=default]
    -offa 0 [=default]
    -offx 0 [=default]
    -offy 0 [=default]
    -alpha 2.00 [=default]
    -beta 2.00 [=default]
    -platform 0 [=default]
    -device 0 [=default]
    -num_steps 0 [=default]
    -runs 10 [=default]
    -h [false]
    -q [false]
    -no_abbrv [false]
    -full_statistics [false]
    -warm_up [false]
    -tuner_files <none> [=default]

                                                                                                                         | <--       CLBlast       --> | <--      CPU BLAS       --> |
        m;        n;   layout;   transA;      lda;     incx;     incy;     offa;     offx;     offy;    alpha;     beta;     ms_1; GFLOPS_1;    GBs_1;     ms_3; GFLOPS_3;    GBs_3
      512;      512;      101;      111;      512;        1;        1;        0;        0;        0;     2.00;     2.00;     0.27;      1.9;      3.9;     0.02;     29.6;     59.6
```

And here is the current master branch:
```
* Options given/available:
    -m 512 [=default]
    -n 512 [=default]
    -layout 101 (row-major) [=default]
    -transA 111 (regular) [=default]
    -lda 512 [=default]
    -incx 1 [=default]
    -incy 1 [=default]
    -offa 0 [=default]
    -offx 0 [=default]
    -offy 0 [=default]
    -alpha 2.00 [=default]
    -beta 2.00 [=default]
    -platform 0 [=default]
    -device 0 [=default]
    -precision 32 (single) [=default]
    -cblas 1 [=default]
    -step 1 [=default]
    -num_steps 0 [=default]
    -runs 10 [=default]
    -h [false]
    -q [false]
    -no_abbrv [false]
    -full_statistics [false]
    -warm_up [false]
    -tuner_files <none> [=default]

                                                                                                                         | <--       CLBlast       --> | <--      CPU BLAS       --> |
        m;        n;   layout;   transA;      lda;     incx;     incy;     offa;     offx;     offy;    alpha;     beta;     ms_1; GFLOPS_1;    GBs_1;     ms_3; GFLOPS_3;    GBs_3
      512;      512;      101;      111;      512;        1;        1;        0;        0;        0;     2.00;     2.00;     0.41;      1.3;      2.6;     0.02;     25.2;     50.7
```

As a side question, why is performance of CLBlast so poor compared to CPU BLAS? My max bandwidth from running the copy kernel is about 20-25 GB/s and yet it achieves about 2.6 GB/s.

CNugteren (approved):
Let's merge this in. Even if there is no speed benefit or almost no speed benefit it can't hurt I guess, and also serves as documentation of the assumptions :+1: I tested it locally and see nothing strange.

PR #620, opened 2025-08-29
Improved Compilation Time

EJain-Dev:
Moved some functions that were in the utility headers into source files to improve compile times. I didn't move the batch functions because they are only included by 2 files.


EJain-Dev:
I started work on a clang tidy PR and this isn't too major of a improvement so I'll close this for now.

PR #621, opened 2025-08-29
Implemented Clang Tidy

EJain-Dev:
I went through and did a lot of clang tidy checks to start integrating this codebase with all of modern features.


EJain-Dev:
I don't think this will be very successful unless scripts are removed, please let me know your opinion on that, and if you want to keep the scripts, I will close this PR.

EJain-Dev:
Not all files have been linted yet, but I think putting the clang tidy file in now will help for future development and the other files (tuning and tests) can also be linted as they are edited.

src/api_common.cpp
```diff
@@ -107,14 +107,14 @@ void FillCacheForPrecision(Queue& queue) {
     Xhemm<Complex>(queue, nullptr);
     Xherk<Complex, Real>(queue, nullptr);
     Xher2k<Complex, Real>(queue, nullptr);
-    Xspr2<Real>(queue, nullptr);
```
CNugteren: Why was this removed?

src/api_common.cpp
```diff
@@ -142,8 +142,7 @@ StatusCode RetrieveParameters(const RawDeviceID device, const std::string& kerne
   try {
     // Retrieves the device name
     const auto device_cpp = Device(device);
-    const auto platform_id = device_cpp.PlatformID();
-    const auto device_name = GetDeviceName(device_cpp);
+    auto* const platform_id = device_cpp.PlatformID();
```
CNugteren: What's wrong with `const auto`?

src/kernel_preprocessor.cpp
```diff
@@ -70,7 +70,7 @@ bool HasOnlyDigits(const std::string& str) {
 // Simple unsigned integer math parser
 int ParseMath(const std::string& str) {
   // Handles brackets
-  if (str.find(")") != std::string::npos) {
+  if (str.find(')') != std::string::npos) {
```
CNugteren: Is this still the same? It used to be a string and now it is a char, right?

CNugteren (commented):
This PR makes a lot of changes. Not sure if I can properly review all of it. Is there a way to split this up per type of change perhaps? Or would that be a lot of work for you?

I just went through 10 out of the 150 files and gave some comments.

src/api_common.cpp
EJainDev: For clarity, it is suggested to use auto* const instead of const auto which has the same effect. Found throughout the code.

src/database/database.cpp
```diff
@@ -221,7 +221,7 @@ database::Parameters Database::Search(const std::string& this_kernel, const std:
   for (auto& db : this_database) {
     if ((db.kernel == this_kernel) && (db.precision == this_precision || db.precision == Precision::kAny)) {
       // Searches for the right vendor and device type, or selects the default if unavailable
-      const auto parameters =
+      auto parameters =
```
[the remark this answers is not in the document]
EJainDev: Const correctness prevents a move, thus potentially preventing compiler optimization.

EJainDev:
I have to fix IWYU on this PR, so can't be merged yet.

CNugteren:
> There is no easy way to simplify the readability of these changes, however, I will point out that most of these are not functional changes. They simply either improve code readability or performance with minimal actual changes. I would say most of the actual changes are regarding the preprocessor file or files that use multiple kernels, because the ternary operator has been replaced with if statements for 2 or more ternary operators.

Thanks, I understand. Yeah indeed ideally most of your changes could likely be blindly accepted. Still, we have seen in other linting PRs in the past (e.g. IWYU) that even those changes can break functionality. So I'm a bit afraid of simply pressing the accept button here, especially after seeing the first few files and already noticing a few issues. I understand that the tests should cover most of the correctness, but speed is another issue that can't be easily checked. And then there is code-style, with which I mean that blindly applying a linter might not always be the best idea. I also can't go and review all 37,221 additions and 25,945 removals. Or better said, I could review them, but then it has to be worth my time.

So I'm not really sure what to do with this PR yet. What do you think?

One option could be to make a new PR where you add clang-tidy checks to CI with the most minimal amount of checks possible (and thus only a very limited set of fixes). Ideally each commit should be one type of change, such that it is easy to review. And then once we merge that in we might add a second PR with a few more checks? I understand that is work for you of course, so feel free to disagree.

src/database/database.cpp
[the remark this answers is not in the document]
EJainDev: https://clang.llvm.org/extra/clang-tidy/checks/performance/no-automatic-move.html

EJainDev:
I think it would be fastest to do multiple PRs, so I’ll leave this as reference for now, but create mini PR’s to build up to this.

PR #624, opened 2025-09-02
Generator script removal

EJain-Dev:
This PR just removes the Github Action for the generator scripts and puts a warning in scripts/generator/generator.py, I kept the scripts for now in case of wanting to bring them back.


.github/workflows/lint.yml
```diff
@@ -17,8 +17,6 @@ jobs:
     runs-on: ubuntu-latest
```
CNugteren, line 17: You can remove this whole `GenerateScript` job, it was only meant to make sure the generator script was updated.

CNugteren (commented):
Can you add a quick line to the changelog regarding this change?

CNugteren (approved):

PR #625, opened 2025-09-02
Improved compile time

EJain-Dev:
This moves a few header-only functions into source file to slightly improve compile time.


EJain-Dev:
It's about a 1 second speed-up (33%), I guess this isn't too much of a speed-up, so I'll close this.

PR #629, opened 2025-09-04
Added explicit templates for every routine

EJainDev:
This adds explicit templates for every routine to suppress the warning that pops up.  (I deleted old branch because of some integration issues).


EJain-Dev:
The flag is `-Wunused-var-template` for clang and it seems this also will improve compile time ([here](https://stackoverflow.com/questions/8130602/using-extern-template-c11-to-avoid-instantiation)).

PR #634, opened 2025-09-07
MinMax

### LINKED ISSUE #568: Find min/max in one function
I need both min/max be found in a vector. But only find clblast::Max() and clblast::Min. Is there one such as clblast::MinMax() to find both min/max in one run?

EJainDev:
This resolves #568 although the test doesn't pass. Any ideas on why it doesn't pass? Most of the kernel implementation is xamax but changed for max and min.

### DIFF
--- a/test/routines/levelx/xminmax.hpp
+++ b/test/routines/levelx/xminmax.hpp
@@ -92,6 +92,36 @@ class TestXminmax {
 
   // Note: No CBlas or CLBlas routine exists to compare against so instead it compares against the results of Max and
   // Min routines in CLBlast.
+  static StatusCode RunReference1(const Arguments<float2>& args, BuffersHost<float2>& buffers, Queue& queue) {
+    auto minmax = std::minmax_element(buffers.x_vec.begin() + args.x_offset,
+                                      buffers.x_vec.begin() + args.x_offset + args.n, [](const T& arg1, const T& arg2) {
+                                        return (std::fabs(arg1.real()) + std::fabs(arg1.imag())) <
+                                               (std::fabs(arg2.real()) + std::fabs(arg2.imag()));
+                                      });
+    buffers.scalar_uint[args.imax_offset] =
+        static_cast<unsigned int>(std::distance(buffers.x_vec.begin() + args.x_offset, minmax.second));
+    buffers.second_scalar_uint[args.imin_offset] =
+        static_cast<unsigned int>(std::distance(buffers.x_vec.begin() + args.x_offset, minmax.first));
+    return StatusCode::kSuccess;
+  }
+
+  // Note: No CBlas or CLBlas routine exists to compare against so instead it compares against the results of Max and
+  // Min routines in CLBlast.
+  static StatusCode RunReference1(const Arguments<double2>& args, BuffersHost<double2>& buffers, Queue& queue) {
+    auto minmax = std::minmax_element(buffers.x_vec.begin() + args.x_offset,
+                                      buffers.x_vec.begin() + args.x_offset + args.n, [](const T& arg1, const T& arg2) {
+                                        return (std::fabs(arg1.real()) + std::fabs(arg1.imag())) <
+                                               (std::fabs(arg2.real()) + std::fabs(arg2.imag()));
+                                      });
+    buffers.scalar_uint[args.imax_offset] =
+        static_cast<unsigned int>(std::distance(buffers.x_vec.begin() + args.x_offset, minmax.second));
+    buffers.second_scalar_uint[args.imin_offset] =
+        static_cast<unsigned int>(std::distance(buffers.x_vec.begin() + args.x_offset, minmax.first));
+    return StatusCode::kSuccess;
+  }
+
+  // Note: No CBlas or CLBlas routine exists to compare against so instead it compares against the results of Max and
+  // Min routines in CLBlast.
   static StatusCode RunReference1(const Arguments<float>& args, BuffersHost<float>& buffers, Queue& queue) {
     auto minmax =
         std::minmax_element(buffers.x_vec.begin() + args.x_offset, buffers.x_vec.begin() + args.x_offset + args.n);
@@ -120,21 +150,6 @@ class TestXminmax {
     auto minmax =
         std::minmax_element(buffers.x_vec.begin() + args.x_offset, buffers.x_vec.begin() + args.x_offset + args.n,
                             [](half arg1, half arg2) { return HalfToFloat(arg1) < HalfToFloat(arg2); });
-    buffers.scalar_uint[args.imax_offset] =
-        static_cast<unsigned int>(std::distance(buffers.x_vec.begin() + args.x_offset, minmax.second));
-    buffers.second_scalar_uint[args.imin_offset] =
-        static_cast<unsigned int>(std::distance(buffers.x_vec.begin() + args.x_offset, minmax.first));
-    return StatusCode::kSuccess;
-  }
-
-  // Note: No CBlas or CLBlas routine exists to compare against so instead it compares against the results of Max and
-  // Min routines in CLBlast.
-  static StatusCode RunReference1(const Arguments<T>& args, BuffersHost<T>& buffers, Queue& queue) {
-    auto minmax = std::minmax_element(buffers.x_vec.begin() + args.x_offset,
-                                      buffers.x_vec.begin() + args.x_offset + args.n, [](const T& arg1, const T& arg2) {
-                                        return (std::fabs(arg1.real()) + std::fabs(arg1.imag())) <
-                                               (std::fabs(arg2.real()) + std::fabs(arg2.imag()));
-                                      });
     buffers.scalar_uint[args.imax_offset] =
         static_cast<unsigned int>(std::distance(buffers.x_vec.begin() + args.x_offset, minmax.second));
     buffers.second_scalar_uint[args.imin_offset] =


EJain-Dev:
I tested a dummy script, it seems to be a problem in the test implementation rather than in the actual code implementation, do you know what the error is?

EJain-Dev:
I did some debugging, and the result doesn't make sense, in the `TestBlas` function for the correctness test, this is the debugging info. where `index` = 0 `result`` is the comparison implementation from Min and Max, `result2` is this algorithm, for some reason Max is finding a lower max than the MinMax routine. Should I move it into a level 4 BLAS routine to disable accuracy check for the tests and use a manual comparison implementation not relying on Min and Max routines?

```
result1[index]
1548.00000

result2[index]
3163.00000

x_source_[1548]
0.659978092-i*0.00783841964
real =
0.659978092
imag =
-0.00783841964

x_source_[3163]
-0.738334417+i*0.730397105
real =
-0.738334417
imag =
0.730397105
```

src/api_common.cpp
```diff
@@ -107,7 +108,6 @@ void FillCacheForPrecision(Queue& queue) {
     Xhemm<Complex>(queue, nullptr);
     Xherk<Complex, Real>(queue, nullptr);
     Xher2k<Complex, Real>(queue, nullptr);
-    Xspr2<Real>(queue, nullptr);
```
CNugteren, line 110: Why is this removed?

src/pyclblast/src/pyclblast.pyx
```diff
@@ -2486,6 +2486,69 @@ def gemmStridedBatched(queue, m, n, k, batch_count, a, b, c, a_ld, b_ld, c_ld, a
         raise RuntimeError("PyCLBlast: 'CLBlastXgemmStridedBatched' failed: %s" % get_status_message(err))
     return cl.Event.from_int_ptr(<ptrdiff_t>event)
 
+####################################################################################################
+# Vector scaling: SMINMAX/DMINMAX/CMINMAX/ZMINMAX/HMINMAX
```
CNugteren, line 2490: `Vector scaling` ?

src/routines/levelx/xminmax.hpp
```diff
@@ -0,0 +1,39 @@
...
+  // Templated-precision implementation of the routine
+  void DoMinmax(const size_t n, const Buffer<unsigned int>& imax_buffer, const size_t imax_offset,
+                const Buffer<unsigned int>& imin_buffer, const size_t imin_offset, const Buffer<T>& x_buffer,
+                const size_t x_offset, const size_t x_inc);
+};
+
+// =================================================================================================
+}  // namespace clblast
+
+// CLBLAST_ROUTINES_XOMATCOPY_H_
```
CNugteren, line 38:
```suggestion
```

test/correctness/misc/override_parameters.cpp
```diff
@@ -7,13 +7,19 @@
...
+#include <cstddef>
+#include <cstdio>
+#include <cstdlib>
 #include <iostream>
 #include <random>
 #include <string>
 #include <unordered_map>
 #include <vector>
 
+#include "clblast.h"
```
CNugteren, line 19: I guess this include should be removed to work with the CUDA API as well?

test/correctness/misc/override_parameters.cpp
```diff
@@ -7,13 +7,19 @@
 //
 // =================================================================================================
 
+#include <cstddef>
+#include <cstdio>
+#include <cstdlib>
```
CNugteren, line 12: I guess this wasn't catched by IWYU before somehow?

src/utilities/utilities.hpp
```diff
@@ -237,6 +240,7 @@ struct Arguments {
   size_t c_size = 1;
   size_t ap_size = 1;
   size_t scalar_size = 1;
+  size_t second_scalar_size = 1;
```
CNugteren, line 243: Perhaps it would have been easier if you designed the API of the new function to just return a single buffer with 2 values, of which the first is the min and the second the max. That would have require a lot less changes in the CLBlast infrastructure (including the tests) and would have made reviewing also a lot easier. But I guess that's too late now...

test/routines/levelx/xminmax.hpp
```diff
@@ -0,0 +1,168 @@
...
+    buffers.scalar_uint.Read(queue, args.scalar_size, result_uint);
+    buffers.second_scalar_uint.Read(queue, args.second_scalar_size, second_result_uint);
+    // The result is an integer. However, since the test infrastructure assumes results of
+    // type 'T' (float/double/float2/double2/half), we store the results into T instead.
+    // The values might then become meaningless, but a comparison for testing should still
+    // be valid to verify correctness.
+    auto result_as_T = static_cast<T>(result_uint[0]);
+    std::vector<T> result(args.scalar_size + args.second_scalar_size);
+    result[0] = result_as_T;
+    result[1] = static_cast<T>(second_result_uint[0]);
```
CNugteren, line 150: Perhaps easier to read?
```suggestion
    std::vector<T> result(args.scalar_size + args.second_scalar_size);
    result[0] = static_cast<T>(result_uint[0]);;
    result[1] = static_cast<T>(second_result_uint[0]);
```

src/kernels/levelx/xminmax.opencl
```diff
@@ -0,0 +1,179 @@
...
+  __local unsigned int iminlm[WGS1];
+
+  const int lid = get_local_id(0);
+  const int wgid = get_group_id(0);
+  const int num_groups = get_num_groups(0);
+
+  // Performs loading and the first steps of the reduction
+  #ifdef ABSOLUTE
+    singlereal max = ZERO;
+    singlereal min = SMALLEST;
```
CNugteren, line 49: I guess this part can be removed as it doesn't do anything?

src/kernels/levelx/xminmax.opencl
```diff
@@ -0,0 +1,179 @@
...
+    const int x_index = id*x_inc + x_offset;
+    #if PRECISION == 3232 || PRECISION == 6464
+      singlereal x = fabs(xgm[x_index].x) + fabs(xgm[x_index].y);
+    #else
+      singlereal x = xgm[x_index];
+    #endif
+
+    #ifdef ABSOLUTE
+      singlereal xmin = -fabs(x);
+      x = fabs(x);
```
CNugteren, line 69: Also this can be removed? Perhaps for clarify of the code you can then remove the `xmin = -x` and just use `-x` below? That way it is symmetric with the max part of the code.

src/kernels/levelx/xminmax.opencl
```diff
@@ -0,0 +1,179 @@
...
+#else
+  __kernel __attribute__((reqd_work_group_size(WGS1, 1, 1)))
+#endif
+void Xminmax(const int n,
+           const __global real* restrict xgm, const int x_offset, const int x_inc,
+           __global real* restrict mgm, __global unsigned int* igm) {
+  __local singlereal maxlm[WGS1];
+  __local singlereal minlm[WGS1];
+  __local unsigned int imaxlm[WGS1];
+  __local unsigned int iminlm[WGS1];
```
CNugteren, line 40: An issue here is that this uses the tuning parameters from the `Xdot` routine, which don't use as much local memory as this routine. And thus this might lead to invalid kernels, depending on the values for WGS1. So I guess you'll need to introduce separate tuning variables as well. Sounds like quite some work, perhaps there is an easier way around this? Simply halving WGS1 here to be safe with local memory size but slightly less optimal for speed? I think that could work.

src/kernels/levelx/xminmax.opencl
```diff
@@ -0,0 +1,179 @@
...
+  __kernel __attribute__((reqd_work_group_size(WGS2, 1, 1)))
+#endif
+void XminmaxEpilogue(const __global singlereal* restrict mgm,
+                   const __global unsigned int* restrict igm,
+                   __global unsigned int* imax, const int imax_offset,
+                   __global unsigned int* imin, const int imin_offset) {
+  __local singlereal maxlm[WGS2];
+  __local singlereal minlm[WGS2];
+  __local unsigned int imaxlm[WGS2];
+  __local unsigned int iminlm[WGS2];
```
CNugteren, line 130: Same issue here as above regarding local memory usage.

CNugteren (commented):
I went through the code and left a few comments. I can't see any obvious clear bugs though, apart from the potential issue with the re-use of tuning parameters.

As for the tests that you did, I assume you testing all of the following:
* Run with the `-full_test` option?
* Run with a variety of values for WGS1 and WGS2?

src/kernels/levelx/xminmax.opencl
EJainDev: Same as above: I have these in for a Xaminmax routine to replicate the structure of Xmin/Xamin after I get this working.

src/utilities/utilities.hpp
EJainDev: I feel that this provides more flexibility but I guess I could have used 1 buffer as the argument for bot imax and Imin, but then that might cause race conditions.

EJainDev:
I think the issue was with local memory. It was using to much local memory and then when I fixed that, WGS1 or WGS2 was not divisible by 2. I will add a Xaminmax version of this, and then I think it will be mergeable.

EJainDev:
I did run both full test and a variety of values for WGS1 and WGS2.

EJainDev:
Not done yet, got to test on another device, and that has a lot of errors. Probably something to do with the testing system...

PR #639, opened 2025-09-17
Fix CUDA API Interportability

EJainDev:
This is a bug fix after the last change of adding proper IWYU. IWYU automatically added #include "clblast.h" to every file, ruining interportability with the CUDA version of clblast_cuda.h, this fixes that.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -343,9 +343,6 @@ target_include_directories(clblast PUBLIC
                            $<INSTALL_INTERFACE:include>
                            ${API_INCLUDE_DIRS})
 
-# Tells CLBlast Includes that they are being used for internal usage
-target_compile_definitions(clblast PRIVATE CLBLAST_BUILDING_CLBLAST)
-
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -343,6 +343,9 @@ target_include_directories(clblast PUBLIC
                            $<INSTALL_INTERFACE:include>
                            ${API_INCLUDE_DIRS})
 
+# Tells CLBlast Includes that they are being used for internal usage
+target_compile_definitions(clblast PRIVATE CLBLAST_BUILDING_CLBLAST)
```
CNugteren, line 347: Hmm, I find this define a bit weird. Can't you just make this PR without that new define, i.e. fix the CUDA API only? And then perhaps in a separate PR we can discuss how to make your linter happy - or - my preference - not at all: your linter isn't automatically applied anyway, so no need to introduce this verbosity and expose it to the user I would say?

CNugteren (approved):

PR #649, opened 2025-11-30
Replaced Custom FindOpenCL.cmake

### LINKED ISSUE #645: Using Builtin FindOpenCL Instead of Manual CMake
Since CMake 3.1, there is a built-in find_package(OpenCL) that can be used instead of the custom .cmake file to find OpenCL, should that be used instead since the minimum by default is already 3.1?

EJainDev:
Resolves #645


CNugteren (changes requested):
The tests are failing. I guess at least these occurrences (also in the docs) need to be updated:
https://github.com/search?q=repo%3ACNugteren%2FCLBlast%20OPENCL_ROOT&type=code

Issue #665, opened 2026-05-24
Tuning issue on directgemm on access memory beyond the largest legal address


octaveoclx:
Tuning on AMD 5700 and received a repeatable bug report:
clblast_tuner_xgemm_direct -precision 3232 -threads 128 -fraction 1
#-threads 1028 produce the same result:

| 2155 |  3347 |   32   32    8   32   32   16    1    1    1    0 |      32       8 |     256      64 |   OK   1376 ms |      0.13 ms |  262.9 |     results match |
| 2156 |  3347 |   32   32    8   32   32   16    1    1    1    1 |      32       8 |     256      64 |   OK   2083 ms |      0.17 ms |  195.3 |     results match |
| 2157 |  3347 |   64    8    8    8    8    2    1    1    0    0 |       8       8 |      32      32 |   OK   1068 ms |Kernel Name: XgemmDirectTN
VGPU=0x55f6e732ba40 SWq=0x7ca914046000, HWq=0x7ca904600000, id=2
	Dispatch Header =0x1502 (type=2, barrier=1, acquire=2, release=2), setup=0
	grid=[32, 32, 1], workgroup=[8, 8, 1]
	private_seg_size=1184, group_seg_size=65536
	kernel_obj=0x7ca73c702ac0, kernarg_address=0x0x7ca8ffa00000
	completion_signal=0x7ca9080edc00, correlation_id=0
	rptr=49616, wptr=49618
:0:rocdevice.cpp            :3956: 0211749989 us:  Callback: Queue 0x7ca904600000 aborting with error : HSA_STATUS_ERROR_MEMORY_APERTURE_VIOLATION: The agent attempted to access memory beyond the largest legal address. code: 0x29
^C

Analysis from DeepSeek:

## Problem Description

A GEMM test running CLBlast (a GPU‑accelerated BLAS library) crashed with the following error on test case #2157:

```
HSA_STATUS_ERROR_MEMORY_APERTURE_VIOLATION: The agent attempted to access memory beyond the largest legal address.
```

**Failing configuration (extracted from the log):**

| Test # | Matrix dimensions (m, n, k) & other params | Grid | Workgroup | Kernel |
|--------|---------------------------------------------|------|-----------|--------|
| 2157   | `64 8 8 8 8 2 1 1 0 0` | `[32, 32, 1]` | `[8, 8, 1]` | `XgemmDirectTN` |

[...]

PR #666, opened 2026-06-05
Refactor matrix loading functions in xgemm_direct_part3 （just for reference)

octaveoclx:
Hi Cedric,
Just used deepseek to modified the file. The modifications have been highlighted with Chinese comments.
I just want to get to your attention to see it this modification is reasonable. 
Currently, I am testing the code to see if it will trigger the problem.
So no merge.


octaveoclx:
still tuning 3232, but it has passed all cases in 16, 32, 64.

octaveoclx:
passed all tuning cases

CNugteren:
Hmm, I'm not sure I follow what's going on here. First of all, your PR seems to be mixing two things: a 'refactor' and adding support for Apple FP16 unconditionally:
* The refactor seems functionally correct but speed-wise it will slow things down unnecessary by using a 'checked' version of the same logic.  So this will slow things down, and we don't gain anything it seems.
* I like the idea of running FP16 on Apple hardware, but I'm not sure we can do it unconditionally, I guess we'll need to check the driver version (as you also said in https://github.com/CNugteren/CLBlast/issues/667) or the hardware itself. I'm not sure which hardware supports this and it might be difficult to say for older hardware that we don't have access to.

octaveoclx:
The first one tries to fix the problem encounter [here](https://github.com/CNugteren/CLBlast/issues/665)
The Apple one (as well as the nvidia one)  was a quick experiment to see the working of FP16. In my other project I used more complicated method to check the version of OpenCL driver. But here I believe it is the user's responsibility to upgrade their old drivers or pick this funcs, otherwise code is messy or new magic never happens.

octaveoclx:
```cpp
    // 默认值：Apple 设备原生支持 half（即使没有扩展），其他设备默认提升到 float
    use_native_half_ = false;
    
    if (user_disabled) {
        // 用户明确禁用原生 half
        use_native_half_ = false;
    } else if (user_enabled) {
        // 用户明确启用原生 half（需要设备支持）
        if (is_apple_device) {
            // Apple 设备肯定支持
            use_native_half_ = true;
        } else if (is_nvidia_device_) {
            // NVIDIA 需要检查驱动版本（可选）
            char driver_version[64];
            clGetDeviceInfo(device, CL_DEVICE_VERSION, sizeof(driver_version), driver_version, nullptr);
            int major = 0, minor = 0;
            sscanf(driver_version, "OpenCL %d.%d", &major, &minor);
            bool driver_new_enough = (major > 3) || (major == 3 && minor >= 0);
            use_native_half_ = driver_new_enough;
        } else {
            // 其他设备尝试检测 cl_khr_fp16 扩展
            use_native_half_ = (extensions.find("cl_khr_fp16") != std::string::npos);
        }
    } else {
        // 环境变量未设置，自动决策
        if (is_apple_device) {
            // Apple Silicon 原生支持 half，默认开启原生模式
            use_native_half_ = true;
        } else {
            // 其他设备默认提升到 float（安全）
            use_native_half_ = false;
        }
    }
```

CNugteren:
> The first one tries to fix the problem encounter https://github.com/CNugteren/CLBlast/issues/665

Keep in mind that the tuner launches kernel with different settings than the actual usage does. In other words, an issue encountered during tuning should not be fixed in the generic way affecting all regular runs. This will need to be fixed on the tuner side itself.

> The Apple one (as well as the nvidia one) was a quick experiment to see the working of FP16. In my other project I used more complicated method to check the version of OpenCL driver

Let's do that in a separate PR, and let's see if there is anything we can do here. Otherwise it might simply break things on certain hardware or older software, which is also not that nice.

PR #670, opened 2026-08-21
Tuning: use nlohmann json instead of manual json

EJainDev:
Replace use of manual JSON insertion with nlohmann json for simplicity. This sets up the baseline for resumable tuning.


CNugteren (changes requested):
Let's not copy a file into this repo, let's instead get it through CMake or something?

EJainDev:
Adjusted. Now `nlohmann_json` is gotten through CMake 3.11 FetchContent.

CNugteren (approved):
Thanks for adding this, looks good!

PR #671, opened 2026-08-21
Tuning: add progressive logging to avoid loss of data from early stop

EJainDev:
An error, crash, or manual stop causes the tuning results to never be written, wasting the time spent. Now, it progressively logs and updates the file after every run so that this data is not lost. Makes it possible to resume tuning with a future PR.


EJainDev:
The lint error, from my research, has to be addressed by a separate PR that disables clang-tidy on headers.

CNugteren:
I guess https://github.com/CNugteren/CLBlast/pull/672 should have been made on top of this one, right? I reviewed that one and then realized this PR is actually a subset of that one. Please see https://github.com/CNugteren/CLBlast/pull/672 for my comments.

EJainDev:
It seems that the changes accidentally got mixed up. I'll close this PR and the other one can be directly merged.

Issue #664, opened 2026-05-17
Changing of Tuning strategy after skipping or driver problem


octaveoclx:
Dear Cedric,

Recently, I was trying to tuning devices I have with -fraction 1 (full settings) for gemm indirect and direct. There seems to be no problem with Apple silicon and better tuning results have been witnessed and will be provided. 
However, for Intel A770 it is another story. During almost 12 hours of tuning on the gmmk = 0 on the non-boundary portion of matrix, the gemm tuner during step 2 started to produce skipping for all the rest of cases on both Linux and Windows platform.  (Frustration comes in.)
This is very time wasting and energy inefficient due to that around 70000 cases have been tried among 120000+cases. But due to the non-response after one case trigger the driver problem, the driver refuses to offer any tuning on the rest of cases. 
Even worse, the tuner will not save any results for the step 2 and will waiting all the rest skipping case to pass (also wasting time and energy) . If I close the program, there is no saving for 70000 cases. If I want to continue, I have to start over again for those 70000 cases if I understand correctly (like I did by switch from Ubuntu to windows)?

Shall someone devise a mechanism to save the result in every case and resume and even skipping certain tuning point if such a problem encountered.

For example,  adding an index item for --fraction = 1 or any case, and also a success flag. Then the user can  open this file and discover from when the results are triggering this problem by also saving . The next time, the user can use command like ...tuner_gemm -fraction 1 -step2 70010 to resume from any point (70010)  to avoid that case. 
```
    {
"index" : 70000,
"success" : NO,
"reason":"l2 skipping", "timeout?" 
....
    },
```
I guess this would help many of us to resume tuning on any devices with driver or register issues.

Best wishes,
Jinchuan

EJainDev:
I think this could be better implemented if a library like `simdjson` was implemented to avoid the overhead of manually adjusting JSON to checkpoint. It provides a single header and source file so it can be added without making a CMake dependency.

EDIT: It can also be trimmed a lot to the necessary feature set

CNugteren:
It looks like a good idea to me but I'm not really sure if it can be implemented easily. If I have time I might have a look, but I can't promise anything.

EJainDev:
Hi @CNugteren, I just created 3 PRs (#670 #671 #672) that collectively address this. I used Github Stacked PRs to create this, but unfortunately you cannot submit a stack to upstream, only to your fork. I can rebase/merge after each PR to fix conflicts if there are any.

[...]

PR #672, opened 2026-08-21
Make Tuning Resumable

EJainDev:
Addresses issue #664  to resume tuning after manual stop/crash.


src/tuning/json_logging.hpp
```diff
@@ -0,0 +1,185 @@
...
+      file_.seekp(0, std::ios::beg);
+      file_.clear();
+      file_.seekp(0, std::ios::beg);
+
+      file_ << json.dump();
+      file_.flush();
+    }
+  }
+
+  void remove_existing_configs(std::vector<Configuration>& configs) {
```
CNugteren: `remove_existing_configs` as a member of the logger sounds weird: we are not removing configs from the logger, but from some other vector. Maybe you can make this a free function and give it a clearer name?

src/tuning/json_logging.hpp
```diff
@@ -0,0 +1,185 @@
...
+          printf(
+              "* WARNING: Currently loaded JSON file does not match the current device and platform. Overwriting the "
+              "existing file.\n");
+        }
+        file_.seekp(0, std::ios::beg);
+        file_.clear();
+        file_.seekp(0, std::ios::beg);
+
+        file_ << json.dump();
+        file_.flush();
```
CNugteren: This is duplicated with the one below, maybe we can avoid the duplication somehow? Also, is this the cleanest way to do this?

src/tuning/json_logging.hpp
```diff
@@ -0,0 +1,185 @@
...
+    json["best_kernel"] = best_kernel.second;
+    json["best_time"] = best_time.second;
+    json["best_parameters"] = best_parameters.second;
+
+    file_.clear();
+    file_.seekp(0, std::ios::beg);
+
+    file_ << json.dump(2);  // Pretty-print the file
+    file_.flush();
+  }
```
CNugteren, line 165: Why do we need to read the file and update the data, and then write it again?

What if we would just read the existing JSON data at the start (or not if we don't set the `resume` option) and then just start adding results as normal and at the end compute the best results - and then dump the entire file? I guess that is more write actions because you constantly write the entire file instead of incremental updates but I don't know if that matters too much for speed. It will certainly make the code a lot cleaner.

CNugteren (commented):
I think the idea is good, thanks for putting in the effort. I left a few comments, will look at the rest after these are addressed.

src/tuning/json_logging.hpp
EJainDev: My main concern with that approach is in regards to massive GEMM tuning as there can be many many combinations and it already has so many parameters that the file could get potentially very large overwrites. Adding to the writing is also the fact that serialization has to happen on a much larger document every time.

As for cleanliness of the code, it would cut out only a few lines, we still have to seek to the beginning meaning it only saves a few lines of code and 2 branches. I can leave a comment if that would help to explain the logic going on here. The `if` block at lines 131-137 above and the 2 line read call here are the ones cut out if we rewrite every time. 

On another note, a small optimization in here could be to use an allocator to avoid repeated memory allocations in `add_tuning_result`.

EJainDev: I could definitely do it that way if you think it would be better for the code hygiene to do full writes every time.

src/tuning/json_logging.hpp
```diff
@@ -0,0 +1,195 @@
+#ifndef CLBLAST_TUNING_JSON_LOGGING_H_
+#define CLBLAST_TUNING_JSON_LOGGING_H_
+
+#include <algorithm>
+#include <cstdio>
+#include <fstream>
+#include <iomanip>
+#include <ios>
+#include <iterator>
+#include <nlohmann/json.hpp>
```
CNugteren, line 10: Why do you think the linter can't find this file? It should run CMake and download the file, right? I don't understand why this is an error.

src/tuning/json_logging.hpp
```diff
@@ -0,0 +1,195 @@
...
+#include "tuning/configurations.hpp"
+#include "tuning/tuning.hpp"
+#include "utilities/backend.hpp"
+#include "utilities/utilities.hpp"
+
+/**
+ * best_kernel
+ * best_time
+ * best_parameters
+ */
```
CNugteren: Was this a temporary comment or? I don't understand what this does here.

src/tuning/json_logging.hpp
```diff
@@ -0,0 +1,195 @@
...
+      if (resume && !old_json.empty() && json["clblast_device_type"] == old_json.at("clblast_device_type") &&
+          json["clblast_device_vendor"] == old_json.at("clblast_device_vendor") &&
+          json["clblast_device_architecture"] == old_json.at("clblast_device_architecture") &&
+          json["clblast_device_name"] == old_json.at("clblast_device_name") &&
+          json["device"] == old_json.at("device") && json["platform_vendor"] == old_json.at("platform_vendor") &&
+          json["platform_version"] == old_json.at("platform_version") &&
+          json["device_vendor"] == old_json.at("device_vendor") && json["device_type"] == old_json.at("device_type") &&
+          json["device_core_clock"] == old_json.at("device_core_clock") &&
+          json["device_compute_units"] == old_json.at("device_compute_units") &&
+          json["device_extra_info"] == old_json.at("device_extra_info")) {
```
CNugteren: Not sure if easy, but just above you already wrote all those names of the files, and here you write them again twice. A bit error prone and not future proof perhaps.

Can we perhaps do something like this instead (pseudo code):
```c++
auto fields = json.keys();
auto all_ok = true;
for (field in fields) {
  if (json[field] != old_json.at(field)) {
    all_ok = false;
    break;
  }
}
if (resume && all_ok) {
   ...
```

src/tuning/json_logging.hpp
```diff
@@ -0,0 +1,195 @@
...
+            for (const auto& param : result["parameters"].items()) {
+              config[param.key()] = param.value();
+            }
+            configs_.insert(config);
+          }
+        } else {
+          old_json["results"] = nlohmann::json::array();
+        }
+
+        file_ << old_json.dump();  // Remove pretty-print for compatibility below
```
CNugteren: Can you elaborate? I tested locally and indeed it seems it is all dumped on a single line. Not super important, but not as human readable perhaps.

src/tuning/json_logging.hpp
CNugteren: Not sure if easy, but just above you already wrote all those names of the files, and here you write them again twice. A bit error prone and not future proof perhaps.

Can we perhaps do something like this instead (pseudo code):
```c++
auto fields = json.keys();
auto all_ok = true;
for (field in fields) {
  if (json[field] != old_json.at(field)) {
    all_ok = false;
    break;
  }
}
if (resume && all_ok) {
   ...
```

src/tuning/json_logging.hpp
```diff
@@ -0,0 +1,195 @@
...
+        file_.flush();
+      } else {
+        if (resume) {
+          printf(
+              "* WARNING: Currently loaded JSON file does not match the current device and platform. Overwriting the "
+              "existing file.\n");
+        }
+        reset_file(file_);
```
CNugteren, line 110: Maybe more user-friendly and easier to just exit in this case?

src/tuning/json_logging.hpp
CNugteren: OK, I see your concern. Let's keep it like it is.

CNugteren (commented):
I also have a comment on your previous PR, which I found when trying to test this locally. It seems that this JSON repository is 150MB or so, which makes it a quite big download, especially given that we are interested in one header file only. Could you change the CMakeLists to download only that header file and not the whole repo? Thanks!

src/tuning/json_logging.hpp
EJainDev: From what I found, it is that clang tidy parses the `json_logging.hpp` file separately. However, there is no compile command that directly uses this file, so the include directory for this file is not available when parsing this. However, when clang tidy parses the source file including the header, it will report any errors.

src/tuning/json_logging.hpp
EJainDev: Done with ignore set for "best_*" and "results" fields.

src/tuning/json_logging.hpp
EJainDev: Aborts now

src/tuning/json_logging.hpp
EJainDev: It is dumped on a single line until the tuning process finishes. Once it finishes, the best kernel evaluation dumps a pretty printed file. This is because `add_tuning_result` specifically goes back for the "]}" characters.

src/tuning/json_logging.hpp
EJainDev: Removed, it was a temporary.

EJainDev:
Hi @CNugteren, I have addressed all feedback.

src/tuning/json_logging.hpp
CNugteren: Hmm but in https://github.com/CNugteren/CLBlast/pull/670 the linter was also happy, so it should be possible to make it happy here as well. Can you reproduce this issue locally? Maybe something needs to be changed in the CMakeLists? Can you look into this?

CNugteren:
Thanks for making the changes.

One more comment. If I run now with `resume -1` and there is no JSON yet the error is quite cryptic:
```
terminate called after throwing an instance of 'nlohmann::json_abi_v3_12_0::detail::parse_error'
  what():  [json.exception.parse_error.101] parse error at line 1, column 1: attempting to parse an empty input; check that your input string or stream contains the expected JSON
```
Perhaps it should just nicely abort instead?

src/tuning/json_logging.hpp
EJainDev: The linter only runs on changed files, and PR #670 didn't change a header file. Looking back at PR #646, the same thing happened where the clang tidy check was failing since the same issue was happening. Only the header file was changed and I am able to reproduce this issue locally.

CNugteren: OK makes sense I guess.

But we should still have a way to prevent this `'nlohmann/json.hpp' file not found` here? Somehow the CI job that runs clang-tidy doesn't download the required file?

